PR #3, opened 2015-10-05
MXKAccount: add block to handle unrecognized/changed certificates dur…

giomfo:
…ing authentication challenge from a server.

### DIFF
--- a/MatrixKit/Categories/NSData+MatrixKit.h
+++ b/MatrixKit/Categories/NSData+MatrixKit.h
@@ -22,17 +22,23 @@
 @interface NSData (MatrixKit)
 
 /**
- Return a MD5 hash code from NSData content
+ Return a MD5 hash code from NSData content.
+ 
+ @return a hex string without space.
  */
 - (NSString*)MD5;
 
 /**
- Return a SHA256 hash code from NSData content
+ Return a SHA256 hash code from NSData content.
+ 
+ @return a hex string without space.
  */
 - (NSString*)SHA256;
 
 /**
- Return a readable SHA256 hash code from NSData content
+ Return a readable SHA256 hash code from NSData content.
+ 
+ @return a hex string with a white space between each value.
  */
 - (NSString*)SHA256AsHexString;
 
--- a/MatrixKit/Categories/NSData+MatrixKit.h
+++ b/MatrixKit/Categories/NSData+MatrixKit.h
@@ -26,4 +26,14 @@
  */
 - (NSString*)MD5;
 
+/**
+ Return a SHA256 hash code from NSData content
+ */
+- (NSString*)SHA256;
+
+/**
+ Return a readable SHA256 hash code from NSData content
+ */
+- (NSString*)SHA256AsHexString;
+
 @end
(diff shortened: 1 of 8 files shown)


MatrixKit/Categories/NSData+MatrixKit.h
```diff
@@ -26,4 +26,14 @@
  */
 - (NSString*)MD5;
 
+/**
+ Return a SHA256 hash code from NSData content
+ */
+- (NSString*)SHA256;
+
+/**
+ Return a readable SHA256 hash code from NSData content
```
dbkr: Could be a little clearer on what the difference is between these two: the first is also a hex string (the second just has spaces)

MatrixKit/Controllers/MXKAuthenticationViewController.m
```diff
@@ -496,6 +378,97 @@ - (void)refreshSupportedAuthFlow
...
+                    isTrusted = YES;
+                    dispatch_semaphore_signal(semaphore);
+                    
+                }];
+                
+                dispatch_async(dispatch_get_main_queue(), ^{
+                    [alert showInViewController:self];
+                });
+                
+                dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
```
dbkr: Mmm, semaphores. Other options to this would be passing a block to the block that it could run when it's made up its mind, or just having a method it runs (which is how the layer below works). That or look into a promises library for Obj-C. This would avoid blocking a thread for the duration the dialog is displayed.

MatrixKit/Models/Account/MXKAccount.h
```diff
@@ -114,6 +125,14 @@ extern NSString *const kMXKAccountErrorDomain;
 @property (nonatomic,getter=isDisabled) BOOL disabled;
 
 /**
+ Register the MXKAccountOnCertificateChange block that will be used to handle certificate change during account use.
+ This block is nil by default, any new certificate is ignored.
```
dbkr: What does 'ignored' mean here?

MatrixKit/Controllers/MXKAuthenticationViewController.m
giomfo: Here the semaphore is used on purpose to block the thread for the duration the dialog is displayed. I want to process the authentication challenge synchronously, and the thread has nothing to do until the user answer.

MatrixKit/Models/Account/MXKAccount.h
giomfo: "ignored" means here "untrusted" by default

MatrixKit/Categories/NSData+MatrixKit.h
giomfo: agree

MatrixKit/Models/Account/MXKAccount.h
dbkr: Okay - I'd probably comment it to say explicitly that it will abort the connection to the server.

MatrixKit/Controllers/MXKAuthenticationViewController.m
dbkr: Yeah, I see. Normally this kind of thing would be done asynchronously so you don't need to block a thread.

dbkr:
LGTM

PR #4, opened 2015-10-13
Read receipts management.

ylecollen:


MatrixKit/Models/Room/MXKRoomDataSource.m
```diff
@@ -1644,16 +1690,14 @@ - (void)processQueuedEvents:(void (^)())onComplete
                         [[NSNotificationCenter defaultCenter] postNotificationName:kMXKRoomDataSourceSyncStatusChanged object:self userInfo:nil];
                     }
                 }
+            
+                [self refreshUnreadCounters:NO];
```
giomfo: You should check whether self has not been destroyed before refreshing unread counters (see the following block)

MatrixKit/Models/Room/MXKRoomBubbleCellData.h
```diff
@@ -65,6 +65,11 @@ typedef enum : NSUInteger {
 @property (nonatomic, readonly) MXSession *mxSession;
 
 /**
+ The room state
+ */
+@property (nonatomic, readonly) MXRoomState *mxRoomState;
```
giomfo: I suggest using the current room state instead of store a snapshot of the room state in each cell data instance.

giomfo:
LGTM

PR #5, opened 2015-10-14
Allow URL scheme of links to be customized

illicitonion:
This will allow Chrome to be set to the default link handler in
applications of they want. And probably other browsers too.

### DIFF
--- a/MatrixKit/Models/MXKAppSettings.m
+++ b/MatrixKit/Models/MXKAppSettings.m
@@ -26,7 +26,7 @@
 static MXKAppSettings *standardAppSettings = nil;
 
 @implementation MXKAppSettings
-@synthesize showAllEventsInRoomHistory, showRedactionsInRoomHistory, showUnsupportedEventsInRoomHistory;
+@synthesize showAllEventsInRoomHistory, showRedactionsInRoomHistory, showUnsupportedEventsInRoomHistory, httpLinkScheme, httpsLinkScheme;
 @synthesize showLeftMembersInRoomMemberList, sortRoomMembersUsingLastSeenTime;
 @synthesize syncLocalContacts, phonebookCountryCode;
 @synthesize presenceColorForOnlineUser, presenceColorForUnavailableUser, presenceColorForOfflineUser;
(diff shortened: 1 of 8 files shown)


giomfo:
Here are 2 remarks on the current implementation. The second point is a blocking point.

1/ You have to set the default value for httpLinkScheme and httpsLinkScheme in [MXKAppSettings init] and [MXKAppSettings reset].

2/  You could not remove the getter of the property 'attributedTextMessage'  in MXKRoomBubbleComponent class. This getter is used to update automatically the text display in chat bubble according to the current event state (We reset on purpose the property 'attributedTextMessage' in 'MXKRoomBubbleCellData' (line 123) on event update.)
For example, with your implementation outgoing messages are frozen in pending display (light gray), they are not updated whereas they are sent (or failed).
Your code relative to NSLinkAttributeName attribute may be added in the existing  'stringAttributesForEvent' method in MXKEventFormatter.m, but I agree that this is not relevant because this attribute does not depend on the event state. So maybe you should add a new property in MXKRoomBubbleComponent class to compute once and store this attribute value.

illicitonion:
On 14/10/15 17:18, giomfo wrote:
> Here are 2 remarks on the current implementation. The second point is a
> blocking point.
>
> 1/ You have to set the default value for httpLinkScheme and
> httpsLinkScheme in [MXKAppSettings init] and [MXKAppSettings reset].

Done

> 2/ You could not remove the getter of the property
> 'attributedTextMessage' in MXKRoomBubbleComponent class. This getter is
> used to update automatically the text display in chat bubble according
> to the current event state (We reset on purpose the property
> 'attributedTextMessage' in 'MXKRoomBubbleCellData' (line 123) on event
> update.)

I re-added getter by way of a @synthesized property, but there's clearly 
some data flow I'm missing here; can you explain how the repaint is 
actually requested? I see the nil-setting, but I'm not sure what is 
meant to then come and set it to a non-nil value.

> For example, with your implementation outgoing messages are frozen in
> pending display (light gray), they are not updated whereas they are sent
> (or failed).
> Your code relative to NSLinkAttributeName attribute may be added in the
> existing 'stringAttributesForEvent' method in MXKEventFormatter.m, but I
> agree that this is not relevant because this attribute does not depend
> on the event state.

I wish :( That would be gloriously simpler. There's no way to add ranges 
to an attribute in the dict, unfortunately.

> So maybe you should add a new property in
> MXKRoomBubbleComponent class to compute once and store this attribute value.

Yeah, I was considering that, but this *feels* like a formatting issue, 
and also the settings are (rightly) available in the formatter but (also 
rightly) not in the RoomBubbleComponent...

giomfo:
About 'attributedTextMessage' getter, I meant that you have to keep the current implementation of this getter (not only re-add @synthesized property) [line 101-118]. A MXKRoomBubbleComponent  instance is composed by only one event. As soon as the event is updated (state change [pending, sent, failed], or redaction) the attributed string is reseted (see [MXKRoomBubbleComponent updateWithEvent:]). Then this attributed string is re-built only when the getter of this property is called.
For example, this getter is called in MXKRoomBubbleCellData (see firstComponent.attributedTextMessage occurrences), or in MXKRoomBubbleCellDataWithAppendingMode (component.attributedTextMessage).

I'm not sure to understood your other remarks. I think you should modify the existing
[MXKEventFormatter stringAttributesForEvent:] in order to return a NSAttributedString (instead of a dictionary of attributes). A new parameter must be added to provide the original string (built during component initialisation, when the actual room state is available).

I suggest:

/**
 Return attributed string representing the event.
 @param event the event.
 @param string The string for the new attributed string.
 */
- (NSAttributedString*) attributedStringForEvent:(MXEvent*)event withOriginalString:(NSString*)string;

In conclusion, your changes related to textMessage and attributedTextMessage must be reverted, and all 'stringAttributesForEvent' occurrences must be replaced by 'attributedStringForEvent'.

illicitonion:
Done, PTAL :)

MatrixKit/Models/MXKAppSettings.m
```diff
@@ -184,6 +193,56 @@ - (void)setShowUnsupportedEventsInRoomHistory:(BOOL)boolValue
     }
 }
 
+- (NSString *)httpLinkScheme
+{
+    if (self == [MXKAppSettings standardAppSettings])
+    {
+        return [[NSUserDefaults standardUserDefaults] stringForKey:@"httpLinkScheme"];
```
giomfo: Return @"http" if no value is defined for the key @"httpLinkScheme".

MatrixKit/Models/MXKAppSettings.m
```diff
@@ -184,6 +193,56 @@ - (void)setShowUnsupportedEventsInRoomHistory:(BOOL)boolValue
...
+    {
+        httpLinkScheme = stringValue;
+    }
+}
+
+- (NSString *)httpsLinkScheme
+{
+    if (self == [MXKAppSettings standardAppSettings])
+    {
+        return [[NSUserDefaults standardUserDefaults] stringForKey:@"httpsLinkScheme"];
```
giomfo: Return @"https" if no value is defined for the key @"httpsLinkScheme".

giomfo:
LGTM. I just add 2 remarks about default values handling for httpLinkScheme and httpsLinkScheme ;)

MatrixKit/Models/MXKAppSettings.m
illicitonion: Done

MatrixKit/Models/MXKAppSettings.m
illicitonion: Done

PR #12, opened 2015-10-20
Bg sync

ylecollen:


MatrixKit/Models/Account/MXKAccount.h
```diff
@@ -125,6 +125,13 @@ typedef BOOL (^MXKAccountOnCertificateChange)(MXKAccount *mxAccount, NSData *cer
 @property (nonatomic,getter=isDisabled) BOOL disabled;
 
 /**
+ Enable the online presence event.
+ 
+ The presence event must not be sent if the application is launched by a push notification.
+ */
+@property (nonatomic) BOOL sendPresenceAfterInit;
```
giomfo: I would prefer name this new property "sendUserPresence"
(NO by default). It should be updated by the application according to the launch conditions.

and use it in [MXKAccount setUserPresence:...] 
line 352: if (mxSession && sendUserPresence) { }

giomfo:
LGTM

PR #13, opened 2015-10-20
Prepare attachments viewer: Add MXKAttachment class.

giomfo:


MatrixKit/Models/Room/MXKAttachment.h
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+#import <Foundation/Foundation.h>
+#import <MatrixSDK/MatrixSDK.h>
+
+/**
+ List attachment types
+ */
+typedef enum : NSUInteger {
+    MXKAttachmentTypeUndefined,
```
manuroe: Does it worth to keep MXKAttachmentTypeUndefined?
It was not used before this refactoring and it is still unused after.

giomfo:
I consider this default value may be useful in the future...

manuroe:
ok. LGTM

PR #15, opened 2015-10-22
Attachment viewer

giomfo:


MatrixKit/Controllers/MXKRoomAttachmentsViewController.m
```diff
@@ -0,0 +1,666 @@
...
+    for (NSUInteger index = 0; index < attachments.count; index++)
+    {
+        MXKAttachment *attachment = attachments[index];
+        
+        if ([attachment.event.eventId isEqualToString:eventId])
+        {
+            currentAttachmentIndex = index;
+        }
+    }
+
```
ylecollen: 1 - currentAttachmentIndex should be updated if the item is found.
2 - a break or a return + reloadData could be done in the if block to same time

PR #19, opened 2015-10-26
Room launch improvement

ylecollen:
-> The member avatars are cached
-> the 0.5s blank screen while opening a room should be removed.


giomfo:
LGTM

PR #22, opened 2015-10-28
MXKRoomDataSource: Added an @autoreleasepool on queued events processing

manuroe:
The added autoreleasepool block reduces this memory usage spike  (which happens when retrieving thousands of events on app resuming):

screen shot 2015-10-27 at 16 59 20


ylecollen:
LGTM

PR #24, opened 2015-10-30
MXSession: V2 sync joining.

giomfo:
Implement joinRoom action in case of server sync v2 use.


manuroe:
LGTM

giomfo:
related to https://github.com/matrix-org/matrix-ios-sdk/pull/28

PR #25, opened 2015-10-30
BugFix SYIOS-170:Public Room: room history is wrong when user joins f…

giomfo:
…or the second time.

The property 'isSync' has been removed from MXRoom object. We consider now the user's membership in room state.

Related to https://github.com/matrix-org/matrix-ios-sdk/pull/31


manuroe:
LGTM

PR #28, opened 2015-11-10
Performance profiling

manuroe:
Did some refactoring according to Xcode profiler.
Tested when doing a big app resuming.


MatrixKit/Models/Room/MXKRoomDataSource.m
```diff
@@ -1790,6 +1804,9 @@ - (void)processQueuedEvents:(void (^)())onComplete
         // Lock on `eventsToProcessSnapshot` to suspend reload or destroy during the process.
         @synchronized(eventsToProcessSnapshot)
         {
+            // Reset the previous snapshot
+        	bubblesSnapshot = nil;
```
giomfo: bubbles snapshot should be reseted after use (see below bubbles= bubblesSnapshot;)

giomfo:
lgtm

PR #30, opened 2015-11-19
Issue: Back failed on attachment view (iOS8).

giomfo:
On iOS < 9: the navigation Bar cannot be hidden on viewWillAppear. We need to wait for viewDidAppear call.


manuroe:
I am ok with the changes

PR #33, opened 2015-11-27
SYIOS-180 Bad scrolling performance on iOS 9

giomfo:
Refactor room bubble table view cells to prevent constraints update when cells are reused.


manuroe:
It would be nice to comment the MXKRoom(Incoming/Outgoing)BubbleTableViewCell classes explaining that they are subclassed in a panel of specialised cells and why (performance)

Any idea (Instruments screenshots, device rotation handling duration or any other measurable figures) of this improvement?

manuroe:
LGTM

PR #36, opened 2015-12-03
MXKDataSource refactoring

giomfo:
Add MXKCollectionViewCell class to define custom UICollectionViewCell.


manuroe:
So far so good

PR #38, opened 2015-12-04
display_room_avatars

ylecollen:
-> display the room members avatar with the Vector style


giomfo:
After discussion, the default vector sender's avatar should be handled by using the new property 'senderAvatarPlaceholder' added in MXKRoomBubbleCellData class
(see matrix-org/matrix-ios-kit@76a9ad7)

PR #39, opened 2015-12-05
Mxk data source refactoring

giomfo:
The aim of this refactoring is to handle the cell view definition at view controller level instead of data source level.

For that:
- we define 2 new methods in MXKDataSourceDelegate protocol: one to retrieve the cell view class adapted to render a cell data, another to retrieve the associated identifier used to dequeue  and reuse table or collection cells.
- we remove the definition of cell view identifiers in data source headers. These identifiers are now handled at view controller level.

### DIFF
--- a/MatrixKit/Controllers/MXKRecentListViewController.m
+++ b/MatrixKit/Controllers/MXKRecentListViewController.m
@@ -15,7 +15,11 @@
  */
 
 #import "MXKRecentListViewController.h"
+
 #import "MXKRoomDataSourceManager.h"
+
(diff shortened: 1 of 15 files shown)


manuroe:
The code and the new logic looks good to me.

1- It would be nice to have a description in this PR to explain (and to record) why we are doing this refactoring.
2- The doc, like https://github.com/matrix-org/matrix-ios-kit/blob/master/README.rst#use-case-1-change-cells-in-the-room-chat,  needs to be updated as some methods have gone.

manuroe:
Go!

PR #40, opened 2015-12-07
Don't delete the pusher when we close the session

dbkr:
only when we logout or disable the account, otherwise we delete and (usually) recreate our pusher every time the app starts, which is a Bad Idea.

Should fix the problem where push turns itself off sometimes and then you get no push until you open the app again.


giomfo:
LGTM - thx Dave

ara4n:
does this also fix the issue where push turns itself off and stays turned off in settings?

giomfo:
ara4n, yes I think so

PR #49, opened 2015-12-18
Customize the table used to retrieve the localized version of a string.

giomfo:
If the key is not defined in this table, the localized string is retrieved from the default table "MatrixKit.strings".


manuroe:
LGTM

PR #51, opened 2015-12-21
mxk_localizedStringForKey: Fixed translation defaulting to MatrixKit translation

manuroe:
The code for "Check first customized table" in mxk_localizedStringForKey did not work.
If a key has no value in the translation file, iOS does not return nil but the key itself.

There is no way on iOS to check value existence :(
The proposed fix is based on the assumption that a developer will not translate the single character string "_". With this implementation, the code does not spend its life in string comparison to check the translation value existence.


giomfo:
lgtm

PR #52, opened 2015-12-22
MXKRoomBubbleTableView: Handle avatar tap

giomfo:
Add member display name in text input when user taps on avatar.
Note: We add here a new event to handle long press on the avatar view.


manuroe:
LGTM

PR #53, opened 2016-01-04
Search api support

manuroe:
Added MXKSearchViewController, a VC to search messages in one room or in all user's rooms.


giomfo:
LGTM

PR #54, opened 2016-01-05
SYIOS-196 - Performance issue in MXKContactManager when resuming the …

giomfo:
…app.

Bug Fix:
- Disable contact refresh whereas the session is not in running state (all the contacts are refreshed when session state becomes running).
- Use the processing queue to encode the resulting contacts dictionary in order to unblock the UI thread.


manuroe:
LGTM

PR #56, opened 2016-01-19
SYIOS-183 Store in-progress messages

manuroe:
Most of the job is done in the sdk. see https://github.com/matrix-org/matrix-ios-sdk/pull/62


giomfo:
LGTM

PR #58, opened 2016-01-25
BF: Outgoing Messages were not stored in case of errors (like 180s timeout)

manuroe:
Remove the messages from the outgoing messages store only when we know there have been successfully sent.


giomfo:
LGTM

PR #61, opened 2016-02-01
Room page loading improvement

manuroe:
The goal of this PR is to remove non required pagination requests when opening a room. It saves network data usage and boost the room content display.

The non required pagination requests occurred:
- the first time the datasource was used. In this case, the room data source made a pagination request instead of using messages from the store - which are often enough to fill the screen.

- when the user came back to a room. When coming back the code made an unexpected pagination request.

This PR depends on https://github.com/matrix-org/matrix-ios-sdk/pull/71.


giomfo:
LGTM

PR #62, opened 2016-02-01
MXKViewController/MXKTableViewController: Improve activity indicator display.

manuroe:
Do not show it if the stopActivityIndicator is called just after (less than 0.3s)


giomfo:
LGTM

PR #64, opened 2016-02-05
Email invitation

manuroe:
Added management of m.room.third_party_invite events.


MatrixKit/Models/Contact/MXKContact.h
```diff
@@ -90,6 +95,15 @@ extern NSString *const kMXKContactMatrixContactPrefixId;
  @param matrixID
  @return MXKContact instance
  */
+- (id)initMatrixContactWithDisplayName:(NSString*)displayName;
+
```
giomfo: Why do you need to add this method?[initMatrixContactWithDisplayName:andMatrixID] may be called with a nil matrix id.
Maybe a comment should be added to the existing method, don't you think so?

PR #65, opened 2016-02-08
BF https://github.com/vector-im/vector-ios/issues/82: Deleted unsent …

manuroe:
…messages keep coming back when the app is relaunched (brought back from backward) after offline mode

https://github.com/vector-im/vector-ios/issues/82


giomfo:
LGTM

PR #66, opened 2016-02-08
Unsuspicious back pagination bug fix

manuroe:
https://github.com/vector-im/vector-ios/issues/79: If messages arrive whilst you are scrolled back, the scroll offset jumps


giomfo:
lgtm

PR #68, opened 2016-02-11
Bug Fix: vector-ios #87

giomfo:
Please can we have default ios long-tap to select and clipboard behaviour again?


manuroe:
LGTM

PR #85, opened 2016-04-18
Added a general mechanism to let cells ask confirmation to the app before doing something

manuroe:
The first usage is the click on a link in the chat history to intercept universal links.


giomfo:
LGTM - the added mechanism is quite complex.

PR #87, opened 2016-04-20
MXKRoomViewController: input tool bar and activies view may be remove…

giomfo:
…d on demand.

The input tool bar view and activities view classes may be defined as nil to remove the current view.


manuroe:
LGTM

PR #96, opened 2016-04-28
MXKRecentTableViewCell: Support user's action on recent cell.

giomfo:
We report the delegate in recent table view cell in order to handle user's actions observed in the cell, like button press or tap gestures.


manuroe:
LGTM

PR #99, opened 2016-04-29
BugFix: Handle the error on joining a room where everyone has left

giomfo:
https://github.com/vector-im/vector-ios/issues/283


manuroe:
LGTM

PR #105, opened 2016-05-13
BF: Application can crash when a video failed to be converted before …

manuroe:
…sending (https://github.com/vector-im/vector-ios/issues/318)

Fixed possible race condition as iOS calls the [AVAssetExportSession exportAsynchronouslyWithCompletionHandler] block on a private thread.


giomfo:
NOK
I succeeded to reproduce the issue with the suggested fix.

To reproduce:
- send a video in chat room A  by selecting a video for which exportAsynchronouslyWithCompletionHandler fails.
- before the failure, go back to Messages and select another room B
- the app crashes when exportAsynchronouslyWithCompletionHandler fails.

giomfo:
lgtm

PR #107, opened 2016-05-17
BugFix: Loading one image thumbnail in a sequence seems to set all fu…

giomfo:
…llres images downloading

https://github.com/vector-im/vector-ios/issues/316


manuroe:
SGTM

PR #108, opened 2016-05-17
BugFix: It's too hard to press names to auto-insert nicks

giomfo:
https://github.com/vector-im/vector-ios/issues/309


manuroe:
LGTM

PR #109, opened 2016-05-17
MXKRoomViewController: Do nothing when clicking on an unsent media.

manuroe:
This issue is related to https://github.com/vector-im/vector-ios/issues/313. It prevents the kit from trying to display an unsent media.


giomfo:
lgtm

Issue #124, opened 2016-06-22
Adding markdown when sending msgs


giomfo:
MatrixKit should support markdown by adding 'formatted_body' and 'format' fields in event content (see [sendTextMessage:success:failure:] in MXKRoomDataSource class).
Note: web client uses external lib: https://github.com/chjj/marked

PR #128, opened 2016-06-28
Use DTCoreText to render HTML

manuroe:
Because the NSAttributedString default HTML renderer uses WebKit on the UI thread. As we need to run in a background task, it created race conditions and the app hang forever.

#124


giomfo:
lgtm

Issue #129, opened 2016-06-28
Manage a whitelist of HTML tags allowed in the display of events with HTML formatted_body


manuroe:
Currently only the img tag is blacklisted.
We should be aligned with vector web that sanitises html like this:
https://github.com/matrix-org/matrix-react-sdk/blob/24223ae2b69debb33fa22fcda5aeba6fa93c93eb/src/HtmlUtils.js#L24

PR #131, opened 2016-06-29
MXEventFormatter: Use the same HTML tags whitelist as matrix-react-sdk

manuroe:
#129
And fixed newlines that were unexpectedly added between bubbles.


giomfo:
lgtm

PR #134, opened 2016-07-04
BF: MXKEventFormatter: MD swallows leading #'s even if there are less than 3

manuroe:
https://github.com/vector-im/vector-ios/issues/423


giomfo:
lgtm

PR #137, opened 2016-07-08
Improve HTML rendering a bit

manuroe:
MXKEventFormatter: Add the defaultCSS property to enrich the default CSS used by DTCoreText.
It fixes the rendering of \<code\> tags: line breaks are kept and a monospace font (Courier) is now used.

There are still 2 issues with DTCoreText:
 - background-color does not work. \<code\> has still no background
 - no left border with \<blockquote\> 

This PR is also the beginning of tests in MatrixKit ;)


giomfo:
lgtm

PR #138, opened 2016-07-08
HTML blockquote is badly rendered: some characters can miss

manuroe:
https://github.com/vector-im/vector-ios/issues/437


giomfo:
lgtm

PR #143, opened 2016-08-12
Vector 443

giomfo:
Redacting membership events should immediately reset the displayname & avatar of room members


manuroe:
LGTM

PR #146, opened 2016-08-19
Bug Fix - Screen still tries to turn off when on a VC

giomfo:
https://github.com/vector-im/vector-ios/issues/521

Set up or teardown the promixity monitoring and enable/disable the idle timer according to call type, state & audio route.
TODO: handle the case where an headset is used.


manuroe:
LGTM

PR #147, opened 2016-08-23
Bug Fix - Screen still tries to turn off when on a VC

giomfo:
https://github.com/vector-im/vector-ios/issues/521

- Handle correctly headset usage


giomfo:
Should fix: vector-im/vector-ios#518

manuroe:
LGTM

PR #148, opened 2016-08-24
BF: When the last message is redacted, [MXKRecentCellData update] mak…

manuroe:
…es paginations loops 
https://github.com/vector-im/vector-ios/issues/520

Use the new MXEventsEnumerator from MatrixSDK to find the right last message to display.


giomfo:
LGTM

PR #153, opened 2016-08-25
Should fix - Sync has got stuck while the app was backgrounded

giomfo:
https://github.com/vector-im/vector-ios/issues/506

According to the logs (attached to the emailed report), the applicationDidEnterBackground was ignored during the session state handling. The session was running whereas the app was backgrounded.


manuroe:
LGTM

PR #159, opened 2016-09-02
Bug Fix - Chat screen: wrong attachment is opened

giomfo:
https://github.com/vector-im/vector-ios/issues/387

Do not force the refresh of the current attachment index on attachments update. The current index value is correct.


manuroe:
SGTM

PR #162, opened 2016-09-07
Until e2e is impl'd, encrypted msgs should be shown in the UI as unencryptable warning text

manuroe:
https://github.com/vector-im/vector-ios/issues/559


giomfo:
OK

PR #163, opened 2016-09-08
Room screen: mention the member name at the cursor position (not a the end)

manuroe:
https://github.com/vector-im/vector-ios/issues/385


giomfo:
LGTM

PR #166, opened 2016-09-08
MXKEventFormatter: Change how the kick reason is displayed

manuroe:
It fixes "kick reason should displayed like the webclient" (https://github.com/vector-im/vector-ios/issues/549)


giomfo:
ok

PR #167, opened 2016-09-09
MXKCallViewController: For 1:1 call, display the other peer informati…

giomfo:
…on instead of the room information.


manuroe:
LGTM

PR #168, opened 2016-09-09
Warn explicitly that your addressbook's emails are uploaded for discovering matrix IDs if you enable that permission

manuroe:
https://github.com/vector-im/vector-ios/issues/551


giomfo:
SGTM

PR #170, opened 2016-09-15
Local contacts: Consider the local contacts when the access is author…

giomfo:
…ized even if the server sync option is disabled.


manuroe:
ok

PR #171, opened 2016-09-16
No ringback tones when placing voice calls in silent mode

manuroe:
https://github.com/vector-im/vector-ios/issues/631


giomfo:
LGTM

PR #177, opened 2016-10-05
Bug Fix - Use contains_url filter for navigating files and showing a …

giomfo:
…file index for a room.

vector-im/vector-ios#652

MXRestClient: Support 3 filtering modes on attachments during the messages search (see `MXMessagesSearchMediaFilter`):
- Default value. No limitation.
- Only the messages which contain an URL are concerned by the search request.
- The messages which contain an URL are not concerned by the search request.

### DIFF
--- a/MatrixKit/Models/Search/MXKSearchDataSource.h
+++ b/MatrixKit/Models/Search/MXKSearchDataSource.h
@@ -75,7 +75,7 @@ extern NSString *const kMXKSearchCellDataIdentifier;
 /**
  Launch a message search homeserver side.
 
- @discussion The result depends on the 'mediaFilter' propertie.
+ @discussion The result depends on the 'roomEventFilter' propertie.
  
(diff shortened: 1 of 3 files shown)


MatrixKit/Models/Search/MXKSearchDataSource.h
```diff
@@ -77,19 +71,11 @@ extern NSString *const kMXKSearchCellDataIdentifier;
...
- @param mxSession the Matrix session to get data from.
- @return the newly created instance.
- */
-- (instancetype)initWithRoomId:(NSString*)roomId andMatrixSession:(MXSession*)mxSession;
 
 /**
  Launch a message search homeserver side.
 
- @discussion The result depends on the 'containsURL' propertie.
+ @discussion The result depends on the 'mediaFilter' propertie.
```
manuroe: 'roomEventFilter' property

PR #191, opened 2016-11-15
Generate thumbnail URL in MXKAttachment

dbkr:
Rather than random things setting the property on it as & when


dbkr:
Oh yes, I forgot the opening brackets. Hard habit to break. :/

PR #194, opened 2016-11-16
Support for displaying encrypted thumbnails

dbkr:
Not the full attachment content yet though

### DIFF
--- a/MatrixKit/Models/Room/MXKAttachment.m
+++ b/MatrixKit/Models/Room/MXKAttachment.m
@@ -206,6 +206,7 @@ static const int kThumbnailHeight = 240;
     if ([[NSFileManager defaultManager] fileExistsAtPath:cacheFilePath]) {
         return [MXKMediaManager loadThroughCacheWithFilePath:cacheFilePath];
     }
+    return nil;
 }
 
 - (void)getThumbnail:(void (^)(UIImage *))onSuccess failure:(void (^)(NSError *error))onFailure {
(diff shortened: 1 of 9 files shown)


MatrixKit/Models/Room/MXKAttachment.m
```diff
@@ -163,13 +185,109 @@ - (NSURL *)getThumbnailUrlForSize:(CGSize)size
...
+    {
+        return thumbnail_file[@"mimetype"];
     }
+    return nil;
+}
+
+- (UIImage *)getCachedThumbnail {
+    NSString *cacheFilePath = [MXKMediaManager cachePathForMediaWithURL:self.thumbnailURL
+                                                                andType:self.thumbnailMimeType
+                                                               inFolder:kMXKMediaManagerDefaultCacheFolder];
```
manuroe: Attachment files should be stored in their room folder so that we delete them when the user leaved the room.

MatrixKit/Models/Room/MXKAttachment.m
```diff
@@ -174,6 +195,19 @@ - (NSString *)thumbnailMimeType
...
+    
+    UIImage *thumb = [MXKMediaManager getFromMemoryCacheWithFilePath:cacheFilePath];
+    if (thumb) return thumb;
+    
+    if ([[NSFileManager defaultManager] fileExistsAtPath:cacheFilePath]) {
+        return [MXKMediaManager loadThroughCacheWithFilePath:cacheFilePath];
+    }
+}
```
manuroe: missing return nil;

PR #204, opened 2016-11-22
Switch to e2e attachments spec v1

dbkr:
* v1 uses 64 bits of the IV rather than all 128. It turns out
   commoncrypto does this anyway and has no option to change it,
   so iOS has always done this, so this is basically adding the
   v: "v1" to the attachment info.
 * Add unit test testing the IV wraparound behaviour (this is what
   tests how many bits of the IV we're using)
* Fix base64 padder
 * Check type of things we get out of the attachment dict to avoid
   crashes due to unknown selectors.


dbkr:
The thinking on not checking the version was that we're anticipating this to be backwards compatible if we were to change it again, so we should try to decode it anyway rather than fail hard (web uses the version field to switch the IV wraparound size, but iOS only supports 64 bit wraparound so we can't check and use 128 if the v field is absent). If we need to make a non-backwards compatible version, we'll put the files under a different json key in the content.

PR #206, opened 2016-11-25
Crypto: Update the display of encrypted events when they are lately decrypted

manuroe:
This is done on kMXEventDidDecryptNotification (https://github.com/matrix-org/matrix-ios-sdk/pull/186)

Port of https://github.com/matrix-org/matrix-react-sdk/pull/556

### DIFF
--- a/MatrixKit/Models/Room/MXKRoomDataSource.m
+++ b/MatrixKit/Models/Room/MXKRoomDataSource.m
@@ -220,6 +220,9 @@ NSString *const kMXKRoomDataSourceTimelineErrorErrorKey = @"kMXKRoomDataSourceTi
             [self onDateTimeFormatUpdate];
             
         }];
+
+        // Listen to events decrypted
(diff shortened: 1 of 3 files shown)


MatrixKit/Models/Room/MXKRoomDataSource.m
```diff
@@ -2382,6 +2387,145 @@ - (void)didMXRoomInitialSynced:(NSNotification *)notif
...
+                    for (NSUInteger i = 1; i < removedEvents.count; i++)
+                    {
+                        MXEvent *removedEvent = removedEvents[i];
+                        if (i == 1)
+                        {
+                            newBubbleData = [[class alloc] initWithEvent:removedEvent andRoomState:self.room.state andRoomDataSource:self];
+                        }
+                        else
+                        {
+                            [newBubbleData addEvent:removedEvent andRoomState:self.room.state];
```
giomfo: Please check here [class instancesRespondToSelector:@selector(addEvent:andRoomState:)]

even if this is not possible that the call does not respond to this selector

PR #210, opened 2016-12-01
Move Media Loader in Matrix SDK

giomfo:
- Modify the dictionary of the progress statistics in order to be able to move media loader at SDK level.


giomfo:
This may be true for kMXKMediaLoaderCompletedBytesCountKey and kMXKMediaLoaderTotalBytesCountKey, but not for kMXKMediaLoaderProgressValueKey.

PR #215, opened 2016-12-10
Fix incorrect language code on iOS 9

VShell:
On iOS 9, preferredLanguages values now include the country code. As a result, they can look like zh-Hans-CN, which is not a valid language code and is longer than 8 characters, causing errors in Synapse's database when using Postgres, which assumes the "lang" field is no longer than that. (SQLite doesn't enforce max lengths on strings, so has no issues aside from the language string being wrong.)

According to [this post](http://stackoverflow.com/a/36012440), this patch should solve the issue by returning zh-Hans instead of zh-Hans-CN, though I have no way of testing it. Testing should be done with an iOS 9 device in Simplified Chinese, against a Synapse running Postgres.

### DIFF
--- a/MatrixKit/Models/Account/MXKAccount.m
+++ b/MatrixKit/Models/Account/MXKAccount.m
@@ -840,7 +840,7 @@ static MXKAccountOnCertificateChange _onCertificateChangeBlock;
                                @"url": self.pushGatewayURL,
                                };
     
-    NSString *deviceLang = [NSLocale preferredLanguages][0];
+    NSString *deviceLang = [[NSLocale currentLocale] objectForKey:NSLocaleLanguageCode];;
     
     NSString * profileTag = [[NSUserDefaults standardUserDefaults] valueForKey:@"pusherProfileTag"];
     if (!profileTag)


MatrixKit/Models/Account/MXKAccount.m
```diff
@@ -840,7 +840,7 @@ - (void)enablePusher:(BOOL)enabled success:(void (^)())success failure:(void (^)
                                @"url": self.pushGatewayURL,
                                };
     
-    NSString *deviceLang = [NSLocale preferredLanguages][0];
+    NSString *deviceLang = [[NSLocale currentLocale] objectForKey:NSLocaleLanguageCode];;
```
giomfo: VShell, the country code may be useful in language code to be able to differentiate for example en-GB and en-US.
I would prefer to keep the original code, and apply your patch only when the retrieved language code is longer than 8 characters. 

Note: please remove the double semicolon.

giomfo:
VShell, I reviewed your patch.

Please see our contributing code: https://github.com/matrix-org/matrix-ios-kit/blob/master/CONTRIBUTING.rst. If you agree to this, then all that's needed is to include the line in your commit or pull request comment:

`Signed-off-by: Your Name <your@email.example.org>`
...using your real name

VShell:
Hey, I think you're right and this'd need to be fixed in Synapse.

PR #217, opened 2016-12-16
Outgoing Messages Refactoring:

giomfo:
- Encryption is now handled by SDK (MXRoom sending API)
- Local echo is handled by MXRoom too.


MatrixKit/Utils/MXKEventFormatter.m
```diff
@@ -1362,6 +1362,8 @@ - (UIColor*)textColorForEvent:(MXEvent*)event
                     textColor = _defaultTextColor;
                 }
                 break;
+            case MXEventSentStateEncrypting:
```
manuroe: Events in encrypting state should be in a another color (green in case of Riot).

PR #220, opened 2017-01-05
MXKContactManager: List the local contacts with methods.

giomfo:
- deprecate localEmailContacts:
- add localContactsWithMethods: to list the local contacts who have methods which can be used to invite them or to discover matrix users.


giomfo:
Related to https://github.com/vector-im/riot-ios/issues/904

PR #227, opened 2017-01-18
MXKContactManager: Ignore @facebook.com email addresses from the resu…

giomfo:
…lts - facebook have discontinued that service...


giomfo:
agree

PR #259, opened 2017-03-22
Attachment viewcontroller interactive animations

aramsargsyan:
I have added animation and interaction controllers, which are reusable. Also added an option to make MXKAttachmentsViewController animatable. This pull request is the first step of https://github.com/vector-im/riot-ios/issues/1064 this issue. Afterwards, the reusable UI Elements created here will be used in Riot iOS to finish the improvement mentioned above. Please review my code.

### DIFF
--- a/MatrixKit/Animators/MXKAttachmentInteractionController.m
+++ b/MatrixKit/Animators/MXKAttachmentInteractionController.m
@@ -1,5 +1,5 @@
 /*
- Copyright 2016 OpenMarket Ltd
+ Copyright 2017 Aram Sargsyan
  
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
--- a/MatrixKit/Animators/MXKAttachmentInteractionController.m
+++ b/MatrixKit/Animators/MXKAttachmentInteractionController.m
@@ -1,0 +1,191 @@
+/*
+ Copyright 2016 OpenMarket Ltd
+ 
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ 
+ http://www.apache.org/licenses/LICENSE-2.0
+ 
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "MXKAttachmentInteractionController.h"
+#import "MXKAttachmentAnimator.h"
+
+@interface MXKAttachmentInteractionController ()
+
+@property (weak, atomic) UIViewController<MXKAttachmentAnimatorDelegate> *viewController;
+@property (weak) UIImageView *originalImageViewReference;
+@property CGRect originalImageViewConvertedFrame;
+
+@property UIImageView *transitioningImageView;
+@property id <UIViewControllerContextTransitioning> transitionContext;
+
+@property CGPoint translation;
+@property CGPoint delta;
+
+@end
+
+@implementation MXKAttachmentInteractionController
+
+#pragma mark - Lifecycle
+
+- (instancetype)initWithViewController:(UIViewController<MXKAttachmentAnimatorDelegate> *)viewController originalImageView:(UIImageView *)imageView convertedFrame:(CGRect)frame
+{
+    self = [super init];
+    if (self) {
+        self.viewController = viewController;
+        self.originalImageViewReference = imageView;
+        self.originalImageViewConvertedFrame = frame;
+        self.interactionInProgress = NO;
+        
+        [self preparePanGestureRecognizerInView:viewController.view];
+    }
+    return self;
+}
+
+#pragma mark - Gesture recognizer
+
+- (void)preparePanGestureRecognizerInView:(UIView *)view
+{
+    UIPanGestureRecognizer *recognizer = [[UIPanGestureRecognizer alloc] initWithTarget:self action:@selector(handleGesture:)];
+    recognizer.minimumNumberOfTouches = 1;
+    recognizer.maximumNumberOfTouches = 3;
+    [view addGestureRecognizer:recognizer];
+}
+
+- (void)handleGesture:(UIPanGestureRecognizer *)recognizer
+{
+    CGPoint translation = [recognizer translationInView:self.viewController.view];
+    self.delta = CGPointMake(translation.x - self.translation.x, translation.y - self.translation.y);
+    self.translation = translation;
+    
+    switch (recognizer.state) {
+        case UIGestureRecognizerStateBegan:
+            
+            self.interactionInProgress = YES;
+            
+            if (self.viewController.navigationController) {
+                [self.viewController.navigationController popViewControllerAnimated:YES];
+            } else {
+                [self.viewController dismissViewControllerAnimated:YES completion:nil];
+            }
+            
+            break;
+            
+        case UIGestureRecognizerStateChanged:
+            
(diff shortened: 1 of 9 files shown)


MatrixKit/Animators/MXKAttachmentInteractionController.m
```diff
@@ -0,0 +1,192 @@
+/*
+ Copyright 2016 OpenMarket Ltd
```
giomfo: Please update this copyright by adding your name:
"Copyright 2017 Your Name"

MatrixKit/Controllers/MXKAttachmentsViewController.h
```diff
@@ -97,4 +103,10 @@ limitations under the License.
  @return a boolean which tells whether some new attachments may be added or not.
  */
 - (BOOL)attachmentsViewController:(MXKAttachmentsViewController*)attachmentsViewController paginateAttachmentBefore:(NSString*)eventId;
+
+@optional
+
+- (void)displayedNewAttachmentWithEventId:(NSString *)eventId;
```
giomfo: Please comment this method.

MatrixKit/Controllers/MXKAttachmentsViewController.m
```diff
@@ -1044,6 +1078,12 @@ - (void)collectionView:(UICollectionView *)collectionView didEndDisplayingCell:(
             }
         }
     }
+  
+    //tell the delegate that a new Attachment has been shown and pass eventId
```
giomfo: I don't think this code is at the right place.
Indeed the current cell is not displayed anymore here.
Maybe you should use collectionView:willDisplayCell: method.

MatrixKit/Controllers/MXKAttachmentsViewController.m
```diff
@@ -1306,4 +1346,71 @@ - (void)documentInteractionControllerDidDismissOpenInMenu:(UIDocumentInteraction
...
+    {
+        return [[MXKAttachmentAnimator alloc] initWithAnimationType:PhotoBrowserZoomOutAnimation sourceViewController:self.sourceViewController];
+    }
+    return nil;
+}
+
+- (id<UIViewControllerInteractiveTransitioning>)interactionControllerForDismissal:(id<UIViewControllerAnimatedTransitioning>)animator
+{
+    //if there is no interaction, just finish the animation using the animationController
+    if (self.interactionController.interactionInProgress)
```
giomfo: The comment seems to be the opposite, isn't it?

MatrixKit/Controllers/MXKRoomViewController.m
```diff
@@ -147,6 +147,11 @@ The identifier of the current event displayed at the bottom of the table (just a
     BOOL restartConnection;
 }
 
+@property NSString *openedAttachmentEventId;
+@property NSString *closedAttachmentEventId;
```
giomfo: Please add comments for these 3 new properties.
To be honest, I don't understand the meaning of `closedAttachmentEventId`.

aramsargsyan:
Thank you giomfo very much for thoroughly reviewing my pull request! I have already pushed my changes done according to your comments.
Indeed, the Attachments delegate call was in the wrong place. I have moved it to the scrollViewDidEndDecelerating method (also got the visible cell from there) instead of collectionView:willDisplayCell: because I need the call when the cell is already displayed. 
Signed-off-by: Aram Sargsyan <aram.sargsyan.1997@gmail.com>

MatrixKit/Controllers/MXKAttachmentsViewController.m
```diff
@@ -1113,6 +1107,14 @@ - (void)scrollViewDidEndDecelerating:(UIScrollView *)scrollView
             [self refreshCurrentVisibleCell];
         }
     }
+    
+    //tell the delegate that a new Attachment has been shown and pass eventId
+    MXKMediaCollectionViewCell* cell = [[self.attachmentsCollection visibleCells] firstObject];
+    NSString *attachmentEventId = ((MXKAttachment *)attachments[cell.tag]).eventId;
```
giomfo: Warning: cell.tag may be -1 here, if the current cell is the one with the spinner displayed during the back pagination (isBackPaginationInProgress == YES).

1- I suggest moving the "new attachment notification" inside the `refreshCurrentVisibleCell` method. Then you will be able to use the actual visible attachment.
2- You should call `displayedNewAttachmentWithEventId` with a nil event id, when the current cell is the pagination spinner ((isBackPaginationInProgress == YES) && (currentVisibleItemIndex == 0))

aramsargsyan:
Point taken. 
Also I think that because the refreshCurrentVisibleCell method is not called when opening the Attachments Viewcontroller, it is right to set the closedAttachmentEventId to openedAttachmentEventId when opening the Attachment to be sure that I will use the right frame when closing it. If not for that line, the closedAttachmentEventId would be nil and not equal to openedAttachmentEventId when the user opened and closed the Attachments controller without scrolling left or right, which is not the desired behavior.

MatrixKit/Controllers/MXKAttachmentsViewController.m
```diff
@@ -465,6 +499,17 @@ - (void)refreshCurrentVisibleCell
     
     [self refreshCurrentVisibleItemIndex];
     
+    // Tell the delegate that a new Attachment has been shown and pass eventId
+    MXKMediaCollectionViewCell* cell = [[self.attachmentsCollection visibleCells] firstObject];
```
giomfo: @aramsargsyan: Instead of using the `visibleCells` array here, I would prefer to handle the delegate call during the currentVisibleItemIndex checking. Here is my suggested code for the`refreshCurrentVisibleCell` method:
```
- (void)refreshCurrentVisibleCell
{
    // In case of attached image, load here the high res image.
    
    [self refreshCurrentVisibleItemIndex];
    
    if (currentVisibleItemIndex == NSNotFound)
    {
        // Tell the delegate that a no attachment is displayed for the moment.
        if ([self.delegate respondsToSelector:@selector(displayedNewAttachmentWithEventId:)])
        {
            [self.delegate displayedNewAttachmentWithEventId:nil];
        }
    }
    else
    {
        NSInteger item = currentVisibleItemIndex;
        if (isBackPaginationInProgress)
        {
            if (item == 0)
            {
                // Tell the delegate that a no attachment is displayed for the moment.
                if ([self.delegate respondsToSelector:@selector(displayedNewAttachmentWithEventId:)])
                {
                    [self.delegate displayedNewAttachmentWithEventId:nil];
                }
				
                return;
            }
            
            item --;
        }
        
        if (item < attachments.count)
        {
            MXKAttachment *attachment = attachments[item];
            NSString *attachmentURL = attachment.actualURL;
            NSString *mimeType = attachment.contentInfo[@"mimetype"];
			
            // Tell the delegate which attachment is currently displayed by using its event id.
            if ([self.delegate respondsToSelector:@selector(displayedNewAttachmentWithEventId:)])
            {
                [self.delegate displayedNewAttachmentWithEventId:attachment.eventId];
            }
            
            // Check attachment type
            if (attachment.type == MXKAttachmentTypeImage && attachmentURL.length && ![mimeType isEqualToString:@"image/gif"])
            {
                [...]
            }
        }
    }
}
```

aramsargsyan:
Yeah, I realize that the code is more comprehensive this way.
Can't wait for the community to test the improvement)

giomfo:
@aramsargsyan: I merged your PR on develop. Thx again for your contribution. 
FYI I updated the Authors.rst file.

I tested your animation, it's very cool. Unfortunately I observed the 3 following issues: #262, #263, #264. I will investigate now 262 and 264.

giomfo:
@aramsargsyan: FYI I'm currently working on the 3 issues. All should be fixed in few hours.

giomfo:
All known issues have been fixed.
Thanks to the interactive animation, we don't need anymore to handle a navigation bar. The attachments viewer is now presented modally.

aramsargsyan:
I have thought about presenting it modally and getting rid of the
navigation bar, however didn't want to make a design decision.
Thanks for approving the PR.

PR #279, opened 2017-05-17
Use bundle ressource to store assets

samuel-gallet:
As recommended by Cocoapods https://guides.cocoapods.org/syntax/podspec.html#resource_bundles

Moreover, in a swift project, the assets are not retrieved if the bundle is not handle this way

### DIFF
--- a/MatrixKit/Categories/NSBundle+MatrixKit.m
+++ b/MatrixKit/Categories/NSBundle+MatrixKit.m
@@ -15,6 +15,7 @@
  */
 
 #import "NSBundle+MatrixKit.h"
+#import "MXKViewController.h"
 
 @implementation NSBundle (MatrixKit)
 
@@ -22,9 +23,9 @@ static NSString *customLocalizedStringTableName = nil;
 
 + (NSBundle*)mxk_assetsBundle
 {
-    NSString *bundleResourcePath = [NSBundle mainBundle].resourcePath;
-    NSString *assetPath = [bundleResourcePath stringByAppendingPathComponent:@"MatrixKitAssets.bundle"];
-    return [NSBundle bundleWithPath:assetPath];
+    NSBundle * bundle = [NSBundle bundleForClass:[MXKViewController class]];
+    NSURL * url = [bundle URLForResource:@"MatrixKit" withExtension:@"bundle"];
+    return [NSBundle bundleWithURL:url];
 }
 
 // use a cache to avoid loading images from file system.
--- a/MatrixKit/Categories/NSBundle+MatrixKit.m
+++ b/MatrixKit/Categories/NSBundle+MatrixKit.m
@@ -23,9 +23,16 @@ static NSString *customLocalizedStringTableName = nil;
 
 + (NSBundle*)mxk_assetsBundle
 {
-    NSBundle * bundle = [NSBundle bundleForClass:[MXKViewController class]];
-    NSURL * url = [bundle URLForResource:@"MatrixKit" withExtension:@"bundle"];
-    return [NSBundle bundleWithURL:url];
+    NSBundle *bundle = [NSBundle bundleForClass:[MXKViewController class]];
+    NSURL *assetsBundleURL = [bundle URLForResource:@"MatrixKit" withExtension:@"bundle"];
+
+    // Check whether MatrixKit is not integrated as a Pod
+    if (!assetsBundleURL)
+    {
+        assetsBundleURL = [bundle URLForResource:@"MatrixKitAssets" withExtension:@"bundle"];
+    }
+
+    return [NSBundle bundleWithURL:assetsBundleURL];
 }
 
 // use a cache to avoid loading images from file system.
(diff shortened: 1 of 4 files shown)


giomfo:
@samuel-gallet thanks for your contribution. Could you please sign off your PR (see https://github.com/matrix-org/matrix-ios-kit/blob/master/CONTRIBUTING.rst) by including the line in a pull request comment:

Signed-off-by: Your Name <your@email.example.org>

samuel-gallet:
Signed-off-by: Your Name samuel.gallet@applidium.com

giomfo:
@samuel-gallet: please edit your last comment to replace "Your Name" string with your actual name

samuel-gallet:
It's already done. Can't you see it?

giomfo:
ok this is fine, thx

MatrixKit/Categories/NSBundle+MatrixKit.m
```diff
@@ -15,16 +15,17 @@
...
 @implementation NSBundle (MatrixKit)
 
 static NSString *customLocalizedStringTableName = nil;
 
 + (NSBundle*)mxk_assetsBundle
 {
-    NSString *bundleResourcePath = [NSBundle mainBundle].resourcePath;
-    NSString *assetPath = [bundleResourcePath stringByAppendingPathComponent:@"MatrixKitAssets.bundle"];
-    return [NSBundle bundleWithPath:assetPath];
+    NSBundle * bundle = [NSBundle bundleForClass:[MXKViewController class]];
```
giomfo: Why do you use here [NSBundle bundleForClass:] instead of using [NSBundle mainBundle]?
I would prefer keep using mainBundle:
`NSBundle * bundle = [NSBundle mainBundle];`

samuel-gallet: With cocoa pods 1.x.x, the pod ressources are not copied in the main bundle, they are in the pod bundle. 
So if you are using cocoapods 1.x.x, you cannot load MatrixKit assets by using mainBundle

giomfo: @samuel-gallet did you test your change by running MatrixKitSample? On my side the app crashes on mxk_assetsBundle call. `[bundle URLForResource:@"MatrixKit" withExtension:@"bundle"]` returns nil... (FYI i'm using cocoapods 1.2.1, I observed that the retrived bundle is the main one)

samuel-gallet: Have you tried with a swift project and declared `use_frameworks!` in the podfile?

giomfo: no, I'm using the existing MatrixKitSample target in MatrixKit.xcworkspace.

samuel-gallet: Just have a look at this closed issue here https://github.com/CocoaPods/CocoaPods/issues/3226
It says that when you are using use_frameworks in your podfile (this is mandatory in swift project), the pod is working as a Framework. The pod bundle is accessible by bundleForClass

samuel-gallet: By the way, in the MatrixKitSample project, MatrixKit is not integrated as a Pod!
That's why the retrieved bundle is the main one.
To prevent the app from crash:
NSBundle * bundle = [NSBundle bundleForClass:[MXKViewController class]];
if (bundle == [NSBunde mainBundle]) {
  return bundle
}
NSURL * url = [bundle URLForResource:@"MatrixKit" withExtension:@"bundle"];
return [NSBundle bundleWithURL:url];

giomfo: I agree that the case where the retrieved bundle is the main one, must be handled differently.
But we should not return the main bundle, we want here the assets bundle.
` if (bundle ==  [NSBundle mainBundle])
{
    NSURL * url = [bundle URLForResource:@"MatrixKitAssets" withExtension:@"bundle"];
    return [NSBundle bundleWithURL:url];
} `

Could you please update your PR?

samuel-gallet:
I 've added a new commit.
Can you merge it?

MatrixKit/Categories/NSBundle+MatrixKit.m
```diff
@@ -15,16 +15,21 @@
...
```
giomfo: Your last suggestion failed when MatrixKit is integrated as a Pod in a non swift project.
The right mxk_assetsBundle implementation should be:
```
+ (NSBundle*)mxk_assetsBundle
{
    NSBundle *bundle = [NSBundle bundleForClass:[MXKViewController class]];
    NSURL *assetsBundleURL = [bundle URLForResource:@"MatrixKit" withExtension:@"bundle"];
    
    // Check whether MatrixKit is not integrated as a Pod
    if (!assetsBundleURL)
    {
        assetsBundleURL = [bundle URLForResource:@"MatrixKitAssets" withExtension:@"bundle"];
    }
    
    return [NSBundle bundleWithURL:assetsBundleURL];
}
```
Please update your PR and test it on your side, this should be ok now ;)

samuel-gallet:
Should be good right now.
Thank you!

PR #293, opened 2017-06-08
Read marker - MXKRoomViewController: Improve bubbles table view display.

giomfo:
When the room view controller displays is based on a timeline with an initial event.
The bubble table view content is scrolled by default to display the top of this event at the center of the screen the first time it appears.
We introduce here a new property `centerBubblesTableViewContentOnTheInitialEventBottom` to let the table view center its content on the bottom part of the event. This will be useful to jump to the last unread message and display the marker view.

### DIFF
--- a/MatrixKit/Controllers/MXKRoomViewController.h
+++ b/MatrixKit/Controllers/MXKRoomViewController.h
@@ -99,8 +99,8 @@ extern NSString *const kCmdChangeRoomTopic;
 @property (nonatomic) BOOL updateRoomReadMarker;
 
 /**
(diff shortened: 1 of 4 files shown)


MatrixKit/Controllers/MXKRoomViewController.h
```diff
@@ -99,6 +99,15 @@ extern NSString *const kCmdChangeRoomTopic;
 @property (nonatomic) BOOL updateRoomReadMarker;
 
 /**
+ When the room view controller displays a room data source based on a timeline with an initial event.
```
manuroe: with an initial event,

PR #306, opened 2017-06-19
MXKSoundPlayer

morozkin:
I tried to make `MXKSoundPlayer` as simple as it could be. It suites for tasks of `MXKCallViewController` and can be improved in the future. There are problems with background music and `MXCallViewController` since `MXKCallViewController` doesn't handle all cases. With my PR there is no such problem. It also fixes [1283](https://github.com/vector-im/riot-ios/issues/1283) and the problem with sound after `MXKCallViewController` is dismissed (this problem is reproduced in the last buid from TestFlight. I mentioned about it to you earlier).

**NOTE:** It works great without CallKit but if you try to add CallKit you will notice that after the call is connected, the sound is gone. I have trying to figure out what the problem is. It's obviously that it's something connected with AVAudioSession but i can't find what is wrong.


morozkin:

PR #310, opened 2017-06-21
Read receipts details

aramsargsyan:


MatrixKit/Views/MXKReceiptSendersContainer.h
```diff
@@ -31,6 +31,17 @@ typedef NS_ENUM(NSInteger, ReadReceiptsAlignment)
...
 
+/*
+ Protocol to provide interface for actions related to the MXKReceiptSendersContainer view
+ */
+@protocol MXKRecieptSendersContainerDelegate <NSObject>
+
+@optional
+
+- (void)didTapReceiptsContainerWithRestClient:(MXRestClient *)restClient RoomMembers:(NSArray *)roomMembers avatars:(NSArray *)avatars recieptDescriptions:(NSArray *)recieptDescriptions;
+
```
morozkin: You used generic annotations for `ReadReceiptsViewController ` and i think it's a good idea to add them for `MXKRecieptSendersContainerDelegate `. Also **R**oomMembers

MatrixKit/Views/ReadReceipts/MXKReadReceiptTableViewCell.h
```diff
@@ -0,0 +1,28 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import <UIKit/UIKit.h>
+#import "MXKImageView.h"
+#import "MXKTableViewCell.h"
```
morozkin: `MXKImageView` can be forward declared

MatrixKit/Views/ReadReceipts/MXKReadReceiptTableViewCell.m
```diff
@@ -0,0 +1,41 @@
...
+#import "MXKReadReceiptTableViewCell.h"
+
+
+@implementation MXKReadReceiptTableViewCell
+
+- (void)awakeFromNib
+{
+    [super awakeFromNib];
+    [self configureAvatarImageView];
+}
```
morozkin: Maybe put code from `configureAvatarImageView` right into `awakeFromNib `? You know, it's just 3 lines of code

MatrixKit/Views/MXKReceiptSendersContainer.h
giomfo: About a protocol for a delegate, a good practice is to return the concerned instance as the first parameter:
`- (void)receiptSendersContainer:(MXKReceiptSendersContainer*)receiptSendersContainer...`

then the delegate is able to use the instance's property (you don't have to pass them in parameters), like the REST client here.

BTW after reviewing your PRs, I regret to say that we will not be able to use a delegate at the MXKReceiptSendersContainer level. As we can see in the usage of this delegate in Riot ios (see RoomDataSource file) this is not easy to define the actual delegate. You decided to use `[AppDelegate theDelegate].masterTabBarController.currentRoomViewController` as delegate but nothing guarantee that the current data source is working for this view controller (there are several room view controller instances in the app, like the one used to display the context of a room message resulting of a search session...)

To sum up, you should remove the definition and the usage of this new delegate. I will expose to you an other solution based on the `- (void)cell:(id<MXKCellRendering>)cell didRecognizeAction:(NSString*)actionIdentifier userInfo:(NSDictionary *)userInfo;` defined in MXKCellRenderingDelegate protocol. Sorry to have missed that point before.

MatrixKit/Views/ReadReceipts/MXKReadReceiptTableViewCell.m
```diff
@@ -0,0 +1,41 @@
...
+
+- (void)configureAvatarImageView
+{
+    self.avatarImageView.layer.cornerRadius = CGRectGetWidth(self.avatarImageView.frame)/2;
```
giomfo: You have to override `- (void)layoutSubviews` to set correctly the corner radius. Set this value during `awakeFromNib` may be too early.

Please use the following code:
`- (void)layoutSubviews
{
    [super layoutSubviews];
    
    if (self.avatarImageView)
    {
        // Round image view
        [self.avatarImageView.layer setCornerRadius:self.pictureView.frame.size.width / 2];
        self.avatarImageView.clipsToBounds = YES;
    }
}`

You should move `self.avatarImageView.enableInMemoryCache = YES;` in `awakeFromNib`

MatrixKit/Views/MXKReceiptSendersContainer.h
giomfo: About a protocol for a delegate, a good practice is to return the concerned instance as the first parameter:
- (void)receiptSendersContainer:(MXKReceiptSendersContainer*)receiptSendersContainer...

then the delegate is able to use the instance's property (you don't have to pass them in parameters), like the REST client here.

BTW after reviewing your PRs, I regret to say that we will not be able to use a delegate at the MXKReceiptSendersContainer level. As we can see in the usage of this delegate in Riot ios (see RoomDataSource file) this is not easy to define the actual delegate. You decided to use [AppDelegate theDelegate].masterTabBarController.currentRoomViewController as delegate but nothing guarantee that the current data source is working for this view controller (there are several room view controller instances in the app, like the one used to display the context of a room message resulting of a search session...)

To sum up, you should remove the definition and the usage of this new delegate. I will expose to you an other solution based on the - (void)cell:(id<MXKCellRendering>)cell didRecognizeAction:(NSString*)actionIdentifier userInfo:(NSDictionary *)userInfo; defined in MXKCellRenderingDelegate protocol. Sorry to have missed that point before.

MatrixKit/Views/ReadReceipts/MXKReadReceiptTableViewCell.xib
```diff
@@ -0,0 +1,89 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
giomfo: Could you please adjust the cell height by using for example 74 (instead of 180)?
For the design of this cell, you should copy the existing ContactTableViewCell.xib (riot-ios) by considering only the following items:`thumbnailView, contactDisplayNameLabel and contactInformationLabel` (Do not rename your properties, your naming is fine). 
You could not define a specific label with a fix string like the "Read" label. We won't be able to translate it. You will have to use an attributed string for the `receiptDescriptionLabel`
(Note: remove the white space before ':': "Read: xxx").

MatrixKit/Views/MXKReceiptSendersContainer.h
```diff
@@ -54,6 +54,11 @@ typedef NS_ENUM(NSInteger, ReadReceiptsAlignment)
  */
 @property (nonatomic) MXRestClient* restClient;
 
+/*
+ The MXSession used to format the events in the details view
+ */
+@property (nonatomic) MXSession *mxSession;
```
giomfo: We don't need the matrix session here. The receipt description will be created in the RR details view controller.

MatrixKit/Views/ReadReceipts/MXKReadReceiptTableViewCell.m
```diff
@@ -0,0 +1,46 @@
...
+
+@implementation MXKReadReceiptTableViewCell
+
+- (void)awakeFromNib
+{
+    [super awakeFromNib];
+    self.avatarImageView.enableInMemoryCache = YES;
+    NSMutableAttributedString *readString = [[NSMutableAttributedString alloc] initWithAttributedString:self.readLabel.attributedText];
+    [readString setAttributes:@{NSFontAttributeName:[UIFont boldSystemFontOfSize:15]} range:NSMakeRange(0, readString.length)];
+    self.readLabel.attributedText = readString;
```
giomfo: I advised you to use an attributed string to display the receipt details in only one UILabel instance (instead of using 2 labels:`readLabel` and `receiptDescriptionLabel`).

If you decide to keep this label named `readLabel` for the constant prefix "Read:", you don't have to use attributed string. And you may define the text font in the xib file.
About the label text value, you have to add a new key-value in the MatrixKit.strings file to handle correctly the translation (for example "receipt_status_read" = "Read:";). Then you use the key to define the label text:
`self.readLabel.text = [NSBundle mxk_localizedStringForKey:@"receipt_status_read"]`.

Personally I would prefer to remove the `readLabel` object. We should let the owner of the MXKReadReceiptTableViewCell instance set an attributed string in the `receiptDescriptionLabel` label to include the receipt status ("Read:" or something else). In that case the attributed string will be handled at the application level (in Riot-ios source code for example). CAUTION: the string translation will be handled at application level (see Vector.strings in Riot-ios ressources).

MatrixKit/Assets/MatrixKitAssets.bundle/en.lproj/MatrixKit.strings
```diff
@@ -270,6 +270,7 @@
 "network_error_not_reachable" = "Please check your network connectivity";
 "user_id_placeholder" = "ex: @bob:homeserver";
 "ssl_homeserver_url" = "Home Server URL: %@";
+"receipt_status_read" = "Read: ";
```
giomfo: Please move this key-value at the application level in Vector.strings
You may group it with `read_receipts_list` key.

MatrixKit/Views/ReadReceipts/MXKReadReceiptTableViewCell.m
```diff
@@ -0,0 +1,43 @@
...
+
+@implementation MXKReadReceiptTableViewCell
+
+- (void)awakeFromNib
+{
+    [super awakeFromNib];
+    self.avatarImageView.enableInMemoryCache = YES;
+}
+
+- (void)layoutSubviews {
```
giomfo: Please add a new line before starting a new block {}. Like you did for the other blocks

PR #313, opened 2017-06-22
Fix issues with sound

morozkin:
I decided to move out `AVAudioSession` deactivation from `stopPlaying` method since we need more granular control on when we need to deactivate `AVAudioSession`. Now AVAudioSession is deactivated when it's 100% no sound to play.

As i told earlier we need to start playing ringback sound for outgoing calls only when the system has succesfully activates `AVAudioSession` otherwise the ringback sound will be gone. So i [decided](https://github.com/matrix-org/matrix-ios-sdk/pull/317) to post a global notification about this event and handle it by setting `isRinging` property to `YES` when all criterias are met.

With both PRs we won't have any issues with the sound. I have tested this code and it works well.

Maybe you have your own mind on how we can handle these situations. I'm glad to hear your suggestions.

### DIFF
--- a/MatrixKit/Controllers/MXKCallViewController.m
+++ b/MatrixKit/Controllers/MXKCallViewController.m
@@ -425,7 +425,7 @@ NSString *const kMXKCallViewControllerBackToAppNotification = @"kMXKCallViewCont
         }
         else
         {
-            [[MXKSoundPlayer sharedInstance] stopPlaying];
+            [[MXKSoundPlayer sharedInstance] stopPlayingWithAudioSessionDeactivation:NO];
         }
         
         _isRinging = isRinging;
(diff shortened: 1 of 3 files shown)


MatrixKit/Utils/MXKSoundPlayer.m
```diff
@@ -74,9 +74,6 @@ - (void)stopPlaying
     {
         [self.audioPlayer stop];
         self.audioPlayer = nil;
-        
-        // Release the audio session to allow resuming of background music app
-        [[AVAudioSession sharedInstance] setActive:NO withOptions:AVAudioSessionSetActiveOptionNotifyOthersOnDeactivation error:nil];
```
manuroe: Can we keep this code in the case where we are not running CallKit?
My intention is to be able to reuse this MXKSoundPlayer (for local notifications sounds, for example).

stopPlaying could have an additional bool parameter like deactiveAudioSession:.

MatrixKit/Controllers/MXKCallViewController.m
```diff
@@ -533,9 +547,16 @@ - (void)call:(MXCall *)call stateDidChange:(MXCallState)state reason:(MXEvent *)
             [localPreviewActivityView startAnimating];
             break;
         case MXCallStateCreateOffer:
-            self.isRinging = YES;
+        {
+            BOOL isCallKitAvailable = [MXCallKitAdapter callKitAvailable] && [MXKAppSettings standardAppSettings].isCallKitEnabled;
```
manuroe: Please add a comment explaining why, in case of CallKit, we delay the ringback to the call of handleAudioSessionActivationNotification.
The explanation in the PR comment is pretty clear and can be c+p here.

MatrixKit/Utils/MXKSoundPlayer.m
morozkin: I thought about additional parameter or completion block

manuroe: ATM, I am more for a parameter because AVAudioSession management will stay in MXKSoundPlayer.
MXCallVC is then only a relay between MXCallKitAdapter and MXKSoundPlayer.

PR #318, opened 2017-06-30
Dutch translation of the strings

nvbln:
Note: I have not compiled the file to test if there weren't any errors.

Signed-off-by: Nathan van Beelen


MatrixKit/Assets/MatrixKitAssets.bundle/nl.lproj
```diff
@@ -0,0 +1,447 @@
+/*
+ Copyright 2015 OpenMarket Ltd
```
manuroe: We use only "Copyright 2017 Vector Creations Ltd" now.

PR #352, opened 2017-08-24
Add support for incoming calls view

morozkin:

### DIFF
--- a/MatrixKit/Controllers/MXKCallViewController.h
+++ b/MatrixKit/Controllers/MXKCallViewController.h
@@ -164,6 +164,16 @@ extern NSString *const kMXKCallViewControllerBackToAppNotification;
 - (void)updateProximityAndSleep;
 
 /**
+ Return a view that is used for informing about incoming call.
+ 
+ @return a view that stands for informing about incoming call or nil
+ 
+ @discussion Subclasses must override this method to provide appropriate for their app view.
+ When this method is called peer and mxCall are valid so you can use them.
+ */
+- (UIView *)createIncomingCallView;
+
+/**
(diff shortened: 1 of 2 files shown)


MatrixKit/Controllers/MXKCallViewController.h
```diff
@@ -164,6 +164,16 @@ extern NSString *const kMXKCallViewControllerBackToAppNotification;
 - (void)updateProximityAndSleep;
 
 /**
+ Return a view that is used for informing about incoming call.
+ 
+ @return a view that stands for informing about incoming call or nil
+ 
+ @discussion Subclasses must override this method to provide appropriate for their app view.
+ When this method is called peer and mxCall are valid so you can use them.
+ */
```
giomfo: Could you please improve this comment? 

/**
Prepare and return the optional view displayed during incoming call notification.
Return nil by default
Subclasses may override this method to provide appropriate for their app view.
When this method is called peer and mxCall are valid so you can use them.
*/

MatrixKit/Controllers/MXKCallViewController.m
```diff
@@ -524,6 +526,54 @@ - (void)call:(MXCall *)call stateDidChange:(MXCallState)state reason:(MXEvent *)
...
             self.isRinging = NO;
             [localPreviewActivityView startAnimating];
+            
+            // Try to show a special view for incoming view
+            if (call.isIncoming && !self.incomingCallView)
+            {
+                UIView *incomingCallView = [self createIncomingCallView];
+                if (incomingCallView)
+                {
+                    self.incomingCallView = incomingCallView;
```
giomfo: @morozkin: Clean this view on `destroy`.

PR #359, opened 2017-09-20
BF: Riot on iOS11 sends images as HEIC format, which nothing else can display

manuroe:
vector-im/riot-ios#1521

Force compression for HEIC images so that the kit creates JPEG images for them.


Biep:
Any progress here?

PR #365, opened 2017-10-08
Add support for horizontal rules & H2 headings made out of en-dashes

spantaleev:
Because iOS 11's Smart Punctuation automatically converts
double-hyphens to longer en-dashes (`—`),
it makes it hard to create horizontal rules and H2 headings
via Markdown.

This patch attempts to replace sequences of any kind of dash,
with regular hyphens.

Signed-off-by: Slavi Pantaleev <s.pantaleev@gmail.com>

---

This is a sort of continuation of matrix-org/matrix-ios-kit#364, which fixes the weird `null` bytes within some en-dash messages.

This one ensures that, for the purposes of Markdown generation, a sequence of en-dashes (or mix of en-dashes and hyphens) is treated like a sequence of regular hyphens, so that `cmark` can do its job.

### DIFF
--- a/MatrixKit/Utils/MXKEventFormatter.m
+++ b/MatrixKit/Utils/MXKEventFormatter.m
@@ -1570,6 +1570,28 @@
 
 - (NSString *)htmlStringFromMarkdownString:(NSString *)markdownString
 {
+    if ([markdownString containsString:@"—"]) {
+        //Because iOS 11's Smart Punctuation feature automatically converts double-hyphens
+        //to longer en-dashes (`—`), it becomes complicated to create
+        //- horizontal rules (`text\n\n---\n\ntext` -> `text<hr />text`)
+        //- H2 headings (`Heading\n--` -> `<h2>Heading</h2>`)
+        //
+        //To fix this: for all strings having an en-dash,
+        //normalize sequences of 2 dashes of any kind (hyphen, en-dash) to 2 regular hyphens.
+        //
+        //This should be safe, because we don't expect regular content to contain
+        //2 mixed-kind dashes one after the other.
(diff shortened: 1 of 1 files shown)


MatrixKit/Utils/MXKEventFormatter.m
```diff
@@ -1570,6 +1570,28 @@ - (UIFont*)fontForEvent:(MXEvent*)event
...
+        //normalize sequences of 2 dashes of any kind (hyphen, en-dash) to 2 regular hyphens.
+        //
+        //This should be safe, because we don't expect regular content to contain
+        //2 mixed-kind dashes one after the other.
+        //If it happens, it's likely an attempt to create a horizontal rule or an H2 heading.
+        //
+        //Ideally, we would only be replacing such sequences when they appear on their own
+        //line (a line containing nothing but dashes of some sort).
+        //But that's a more complicated replacement regex for not much gain.
+        NSString *regexFind = @"(-|—){2}";
```
giomfo: @spantaleev: I don't understand how do you fix horizontal rules support with this regex. It works when the user types "——-", but this seems not efficient.

I would prefer a solution where we consider only the lines which contain nothing but dashes of some sort. In these lines, each occurrence of en-dash (`—`) should be replaced with a double-hyphens.
What do you think?

spantaleev: Thinking about it, if the "process only lines with dashes approach" is to be done, it should also make sure to handle lines within quotes, such as:

```
> something
> 
> ---------
> 
> another
```

Probably within multiple levels of quoting as well.

giomfo:
close this outdated PR, a new issue has been filed to keep tracking the request

PR #367, opened 2017-10-12
Do not strip paragraph spacing

spantaleev:
This is a continuation of #355 and aims to fix
the rendering of paragraphs.
(related to vector-im/riot-ios#1500)

Stripping paragraph spacing did not work correctly
and caused content like `<p>first</p><p>second</p>`,
which we expected to render as:

```
first

second
```

to actually render as:

```
first
second
```

This completely defeated the utility of paragraphs and turned
longer messages (those having paragraphs) into long
unreadable walls of text.

The code that used to adjust the spacing of paragraph is removed
with this patch, with the hope that detecting and wrapping
bare HTML messages (like `something<br />another`) prevents
them from being turned into (unwanted) paragraphs.

Since all our paragraphs are now wanted, we can let the spacing be
there for all of them.

The only spacing that we want to remove is the bottom-spacing
of the last paragraph, which we still do.

Signed-off-by: Slavi Pantaleev <s.pantaleev@gmail.com>

### DIFF
--- a/MatrixKit/Utils/MXKEventFormatter.m
+++ b/MatrixKit/Utils/MXKEventFormatter.m
@@ -1154,6 +1154,15 @@
     // Do some sanitisation before rendering the string
     NSString *html = [self sanitiseHTML:htmlString];
 
+    // HTML which is bare, like `something` or `something<br />another`,
+    // would (for some reason) be represented in the NSAttributedString below as a paragraph.
+    // Since our regular `<p>` paragraphs have bottom-spacing (as is expected of paragraphs),
+    // but such bare messages shouldn't, we avoid the paragraph-representation
+    // by wrapping the plain text in another container element.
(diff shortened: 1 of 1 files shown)


giomfo:
@spantaleev: Thanks to your PR, I investigated again vector-im/riot-ios#1500.

In your PR, you enabled spacing in all paragraph style. This fixed your use case: 
screen shot 2017-10-18 at 16 35 25

But this introduced some unwanted space in a block of messages.
(see the space between "Line2" and "Hello" in the following screenshot) 

screen shot 2017-10-18 at 16 35 38

By taking into account your remark, I fixed the issue with https://github.com/matrix-org/matrix-ios-kit/pull/370. This last change should fix vector-im/riot-ios#1500, so I close this PR.
Thanks again for your help

spantaleev:
Glad to know it's fixed now. Thank you!

PR #371, opened 2017-10-22
Add generic annotations for MXKAccountManager

morozkin:
One of the useful things that comes with this improvement is that we will be able to write  `[MXKAccountManager sharedManager].accounts.firstObject.mxRestClient` instead of `[[MXKAccountManager sharedManager].accounts.firstObject mxRestClient]`

### DIFF
--- a/MatrixKit/Models/Account/MXKAccountManager.m
+++ b/MatrixKit/Models/Account/MXKAccountManager.m
@@ -269,7 +269,7 @@ NSString *const kMXKAccountManagerDidRemoveAccountNotification = @"kMXKAccountMa
             [activeAccounts addObject:account];
         }
     }
-    return [activeAccounts copy];
+    return activeAccounts;
 }
 
 - (NSData *)apnsDeviceToken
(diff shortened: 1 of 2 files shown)


MatrixKit/Models/Account/MXKAccountManager.m
```diff
@@ -269,7 +269,7 @@ - (NSArray *)activeAccounts
             [activeAccounts addObject:account];
         }
     }
-    return activeAccounts;
+    return [activeAccounts copy];
```
giomfo: @morozkin Why do you copy here the new created array?

morozkin: @giomfo We have a contract that we will return NSArray and I think it’s better to follow it rather than returning NSMutableArray

giomfo: NSMutableArray inherits from NSArray, so we fulfil the contract ;). I would prefer to remove this copy...

morozkin: It is true, but it's also possible to cast returned array to mutable one and start calling mutating methods. It doesn't look like a very realistic example but there is a possibility of doing such kind of things. I will remove this copy

PR #387, opened 2017-12-27
Missing Push Notifications: Store MXKAccount data (like user credentials) 

manuroe:
into a file instead of NSUserDefaults.

This data should be then more accessible for background processing like handling push notifications.

https://github.com/vector-im/riot-ios/issues/1696

### DIFF
--- a/MatrixKit/Models/Account/MXKAccountManager.m
+++ b/MatrixKit/Models/Account/MXKAccountManager.m
@@ -83,18 +83,11 @@ NSString *const kMXKAccountManagerDidRemoveAccountNotification = @"kMXKAccountMa
 
 - (void)saveAccounts
 {
-    NSUserDefaults *userDefaults = [MXKAppSettings standardAppSettings].sharedUserDefaults;
-    if (mxAccounts.count)
-    {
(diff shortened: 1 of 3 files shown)


MatrixKit/Models/Account/MXKAccountManager.m
```diff
@@ -493,37 +487,64 @@ - (BOOL)isPushAvailable
...
+            NSData *oldAccountData = [[NSUserDefaults standardUserDefaults] objectForKey:kMXKAccountsKey];
+            if (oldAccountData)
+            {
+                [sharedDefaults setObject:oldAccountData forKey:kMXKAccountsKey];
+                [sharedDefaults synchronize];;
 
-            // TODO: Erase old location of accountData when the app goes to the app store
-            //[[NSUserDefaults standardUserDefaults] removeObjectForKey:kMXKAccountsKey];
-            //[[NSUserDefaults standardUserDefaults] synchronize];
+                NSLog(@"[MXKAccountManager] loadAccounts: performed data migration #1");
```
giomfo: Why do you store here this old account in shared Defaults?
I would only check old account storage if none is available in shared defaults:
```
if (!accountData)
{
   accountData = [[NSUserDefaults standardUserDefaults] objectForKey:kMXKAccountsKey];
}
```

PR #394, opened 2018-02-01
BF: iPhone X: room messages overlap the room activity view

manuroe:
(https://github.com/vector-im/riot-ios/issues/1754)


ara4n:
it's weird - i hadn't spotted this at all on my iphone X

PR #395, opened 2018-02-04
Groups: Display flair for users in room history.

giomfo:
MXKRoomDataSource: Force a refresh on flair settings change.
MXKRoomBubbleCellData: keep update the flair of the sender.
MXKRoomBubbleTableViewCell: Display the sender flair if any

TODO: display the full flair list when the user clicks on the displayed flair.

vector-im/riot-meta#118

### DIFF
--- a/MatrixKit/Models/Room/MXKRoomBubbleCellData.m
+++ b/MatrixKit/Models/Room/MXKRoomBubbleCellData.m
@@ -25,7 +25,7 @@
 #import "MXMediaManager.h"
 
 @implementation MXKRoomBubbleCellData
-@synthesize senderId, roomId, senderDisplayName, senderAvatarUrl, senderAvatarPlaceholder, isPaginationFirstBubble, shouldHideSenderInformation, date, isIncoming, isAttachmentWithThumbnail, isAttachmentWithIcon, attachment;
+@synthesize senderId, roomId, senderDisplayName, senderAvatarUrl, senderAvatarPlaceholder, isPaginationFirstBubble, shouldHideSenderInformation, date, isIncoming, isAttachmentWithThumbnail, isAttachmentWithIcon, attachment, senderFlair;
 @synthesize textMessage, attributedTextMessage;
 @synthesize shouldHideSenderName, isTyping, showBubbleDateTime, showBubbleReceipts, useCustomDateTimeLabel, useCustomReceipts, useCustomUnsentButton, hasNoDisplay;
 @synthesize tag;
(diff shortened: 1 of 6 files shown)


MatrixKit/Models/Room/MXKRoomBubbleCellData.m
```diff
@@ -348,6 +351,73 @@ - (void)highlightPatternInTextMessage:(NSString*)pattern withForegroundColor:(UI
...
+                self.senderFlair = flair;
+            }
+            else
+            {
+                self.senderFlair = nil;
+            }
+        }
+        
+        // Observe any change on publicised groups for the message sender
+        [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(didMXSessionUpdatePublicisedGroupsForUsers:) name:kMXSessionDidUpdatePublicisedGroupsForUsersNotification object:self.mxSession];
```
manuroe: Is there a mechanism to refresh the sender flair if the room has been added to a group while the app was in background?
I ask for this particular use case because of the group API design, we only update group data only when the app comes back to foreground.

MatrixKit/Models/Room/MXKRoomDataSource.m
```diff
@@ -606,6 +624,16 @@ - (void)refreshEventListeners:(NSArray *)liveEventTypesFilterForMessages
...
             }
         }];
+        
+        // Flair handling: register a listener for the related groups state event in this room.
+        relatedGroupsListener = [_timeline listenToEventsOfTypes:@[kMXEventTypeStringRoomRelatedGroups] onEvent:^(MXEvent *event, MXTimelineDirection direction, MXRoomState *roomState) {
+            
+            if (MXTimelineDirectionForwards == direction)
+            {
+                // The flair settings have been updated: flush the current bubble data and rebuild them.
+                [self reload];
```
manuroe: is it the answer to my question in the other comment?

MatrixKit/Models/Room/MXKRoomBubbleCellData.m
giomfo: @manuroe: the rooms added to a group are ignored for the moment in the flair process.
Flair takes into account only the related groups defined in room settings and the publicised groups in the user's profile.

MatrixKit/Models/Room/MXKRoomDataSource.m
giomfo: Not really, here we detect that the related groups have been updated. We reload the data source because all the room bubbles have to update their sender flair.

PR #443, opened 2018-06-27
MXKContactManager : Reduce time spent on main thread

SBiOSoftWhare:
Reduce time spent on main thread by MXKContactManager.

Contribute to https://github.com/vector-im/riot-ios/issues/1820

### DIFF
--- a/MatrixKit/Models/Contact/MXKContactField.m
+++ b/MatrixKit/Models/Contact/MXKContactField.m
@@ -51,11 +51,6 @@
     return self;
 }
 
-- (void)dealloc
-{
-    [[NSNotificationCenter defaultCenter] removeObserver:self];
-}
-
 - (void)resetMatrixAvatar
 {
     _avatarImage = nil;
@@ -113,12 +108,17 @@
                 if (!user)
                 {
                     MXSession *mxSession = mxSessions.firstObject;
+                    
+                    MXWeakify(self);
+                    
                     [mxSession.matrixRestClient avatarUrlForUser:_matrixID
                                                          success:^(NSString *mxAvatarUrl) {
                                                              
-                        _matrixAvatarURL = mxAvatarUrl;
+                        MXStrongifyAndReturnIfNil(self);
+                                                             
+                        self.matrixAvatarURL = mxAvatarUrl;
                         
-                        avatarURL = [mxSession.matrixRestClient urlOfContentThumbnail:_matrixAvatarURL toFitViewSize:avatarSize withMethod:MXThumbnailingMethodCrop];
+                        self->avatarURL = [mxSession.matrixRestClient urlOfContentThumbnail:self.matrixAvatarURL toFitViewSize:avatarSize withMethod:MXThumbnailingMethodCrop];
                                                              
                         [self downloadAvatarImage];
                                                              
@@ -146,8 +146,14 @@
         // the image is already in the cache
         if (_avatarImage)
         {
+            __weak typeof(self) weakSelf = self;
+            
             dispatch_async(dispatch_get_main_queue(), ^{
-                [[NSNotificationCenter defaultCenter] postNotificationName:kMXKContactThumbnailUpdateNotification object:_contactID userInfo:nil];
+                
+                if (weakSelf)
+                {
+                    [[NSNotificationCenter defaultCenter] postNotificationName:kMXKContactThumbnailUpdateNotification object:weakSelf.contactID userInfo:nil];
+                }
             });
         }
         else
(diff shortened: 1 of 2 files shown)


MatrixKit/Models/Contact/MXKContactField.m
```diff
@@ -146,8 +146,14 @@ - (void)downloadAvatarImage
         // the image is already in the cache
         if (_avatarImage)
         {
+            __weak typeof(self) weakSelf = self;
```
manuroe: It would be nice to use MXWeakify everywhere now.

MatrixKit/Models/Contact/MXKContactManager.m
```diff
@@ -1230,74 +1229,125 @@ + (void)requestUserConfirmationForLocalContactsSyncWithTitle:(NSString*)title
...
-        for (MXSession *mxSession in mxSessions)
-        {
-            // Check all existing users
-            NSArray *mxUsers = mxSession.users;
+        MXWeakify(self);
+
+        BOOL shouldFetchLocalContacts = self->matrixContactByContactID == nil;
 
-            for (MXUser *user in mxUsers)
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
manuroe: Why do not use the existing `processingQueue` queue?
I am not sure multiplying queues, threads and thread context switches helps here. If I am wrong, we should apply it everywhere.

MatrixKit/Models/Contact/MXKContactManager.m
```diff
@@ -1230,74 +1229,125 @@ + (void)requestUserConfirmationForLocalContactsSyncWithTitle:(NSString*)title
...
+                        contact = [[MXKContact alloc] initMatrixContactWithDisplayName:((user.displayname.length > 0) ? user.displayname : user.userId) andMatrixID:user.userId];
+                    }
+                    
+                    matrixContactByMatrixID[user.userId] = contact;
+                }
+            }
+        }
+    }
+    
+    return [matrixContactByMatrixID copy];
```
manuroe: Do we really need a copy?

MatrixKit/Models/Contact/MXKContactManager.m
```diff
@@ -1230,74 +1229,125 @@ + (void)requestUserConfirmationForLocalContactsSyncWithTitle:(NSString*)title
...
 #pragma mark - Internals
 
+- (NSDictionary*)matrixContactsByMatrixIDFromMXSessions:(NSArray<MXSession*>*)mxSessions
+{
+    // The existing dictionary of contacts will be replaced by this one
+    NSMutableDictionary *matrixContactByMatrixID = [[NSMutableDictionary alloc] init];
+    for (MXSession *mxSession in mxSessions)
+    {
+        // Check all existing users
+        NSArray *mxUsers = mxSession.users;
```
manuroe: I am a bit afraid about accessing this from another thread than the main thread as there is no guarantee that the operation is thread safe.

MatrixKit/Models/Contact/MXKContactManager.m
```diff
@@ -1507,21 +1557,20 @@ - (void)cacheMatrixContacts
...
     NSFileManager *fileManager = [[NSFileManager alloc] init];
     
-    __block NSDictionary *matrixContactByContactIDCpy = nil;
+    __block NSDictionary *matrixContactByContactID = nil;
     
     if ([fileManager fileExistsAtPath:dataFilePath])
     {
         // Use here processing queue because this queue is used during encoding.
         dispatch_sync(processingQueue, ^{
-
```
manuroe: The evil `dispatch_sync` is no more required as we assume that this method is now called from a processing queue.
If not, we should make this method async.

MatrixKit/Models/Contact/MXKContactManager.m
SBiOSoftWhare: It allows to convert NSMutableDictionary to NSDictionary. `copy` always returns an immutable type and `mutableCopy` returns a mutable type. Equivalent to `return [NSDictionary dictionaryWithDictionary: matrixContactByMatrixID];`

MatrixKit/Models/Contact/MXKContactManager.m
SBiOSoftWhare: I agree to get rid of `dispatch_sync` to avoid potential deadlock

MatrixKit/Models/Contact/MXKContactManager.m
SBiOSoftWhare: `dispatch_get_main_queue` gave us a system-defined "background" concurrent queue. Currently we slow down the execution by using the same serial queue for every instructions. But we could use it for the moment to avoid possible race conditions.

MatrixKit/Models/Contact/MXKContactManager.m
SBiOSoftWhare: Yes I could make a copy of MXSession users

PR #481, opened 2018-10-15
E2E export: Enter passphrase -> Create passphrase (#2072)

aaronraimist:
Fixes https://github.com/vector-im/riot-ios/issues/2072


aaronraimist:
I assume I don't need to create a string for every language https://matrix.to/#/!tDRGDwZwQnlkowsjsm:matrix.org/$1539643301101PwodH:raim.ist

manuroe:
> I assume I don't need to create a string for every language 

Yep. I have just git pulled on weblate so that people can translate it in their language: https://translate.riot.im/projects/riot-ios.

PR #508, opened 2019-01-07
BF: Chat screen: wrong thumbnail observed during scrollback

manuroe:
fixes vector-im/riot-ios/issues/1122

Fix a race condition where the image download completes whereas the cell has been reused for another image.

### DIFF
--- a/MatrixKit/Views/MXKImageView.m
+++ b/MatrixKit/Views/MXKImageView.m
@@ -55,6 +55,9 @@
     
     // Subviews
     UIScrollView *scrollView;
+
+    // Current attachment being displayed in the MXKImageView
+    MXKAttachment *currentAttachment;
 }
 @end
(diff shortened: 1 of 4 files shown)


MatrixKit/Views/MXKImageView.m
```diff
@@ -55,6 +55,9 @@ @interface MXKImageView ()
     
     // Subviews
     UIScrollView *scrollView;
+
+    // Current attachment being displayed in the MXKImageView
+    MXKAttachment *currentAttachment;
```
giomfo: Note: `currentAttachment` must be reseted on [setImageURI: withType: andImageOrientation: isThumbnail:: previewImage: mediaManager:] call.
Suggestion: reset `currentAttachment` on [setImage:] call.

Issue #513, opened 2019-02-13
Some japanese strings have a full width percent sign instead of a normal one


neutralinsomniac:
This causes string replacement placeholders (%@) to not get replaced with the relevant values, causing output like this:
![screenshot from 2019-02-13 08-25-03](https://user-images.githubusercontent.com/92071/52717706-af463500-2f6f-11e9-9be3-33036e3bfaf6.png)
![screenshot from 2019-02-13 08-25-23](https://user-images.githubusercontent.com/92071/52717712-b1a88f00-2f6f-11e9-91df-1f0720a9735f.png)

PR #540, opened 2019-04-16
Fix remaining full-width '%' symbols in strings

neutralinsomniac:
Self-explanatory? :)

Signed-off-by: Jeremy O'Brien <neutral@fastmail.com>


manuroe:
Thanks @neutralinsomniac. We will merge it as soon as we can.

neutralinsomniac:
Fixes #513, btw

PR #543, opened 2019-04-23
MXKRoomViewController: remove the implicit retains of "self".

giomfo:
TODO remove the `destroy` method by moving its contain in `dealloc`.

### Pull Request Checklist

* [ ] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

### DIFF
--- a/MatrixKit/Controllers/MXKRoomViewController.m
+++ b/MatrixKit/Controllers/MXKRoomViewController.m
@@ -144,17 +144,17 @@
 /**
  Observe kMXSessionWillLeaveRoomNotification to be notified if the user leaves the current room.
  */
-@property (nonatomic, weak) id kMXSessionWillLeaveRoomNotificationObserver;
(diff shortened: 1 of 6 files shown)


MatrixKit/Controllers/MXKRoomViewController.m
```diff
@@ -152,6 +141,21 @@ @interface MXKRoomViewController ()
 
 @property (nonatomic) UIImageView *openedAttachmentImageView;
 
+/**
+ Observe kMXSessionWillLeaveRoomNotification to be notified if the user leaves the current room.
+ */
+@property (nonatomic, weak) id kMXSessionWillLeaveRoomNotificationObserver;
```
SBiOSoftWhare: It's the opportunity to rename this property. Remove "k" prefix as it is not a constant. And start with lower case: `mxSessionWillLeaveRoomNotificationObserver`. Could be interesting to do the same for the two properties below.

PR #557, opened 2019-05-21
WIP: Switch to UIImagePickerController

aaronraimist:
Works with photos from library but not with photos taken with the camera.

### Pull Request Checklist

* [ ] Pull request is based on the develop branch
* [ ] Pull request updates [CHANGES.rst](https://github.com/matrix-org/matrix-ios-kit/blob/develop/CHANGES.rst)
* [ ] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

Part of: https://github.com/vector-im/riot-ios/pull/2457

### DIFF
--- a/MatrixKit/Views/RoomInputToolbar/MXKRoomInputToolbarView.m
+++ b/MatrixKit/Views/RoomInputToolbar/MXKRoomInputToolbarView.m
@@ -334,7 +334,7 @@
         optionsListView = nil;
     }
     
-    [self dismissMediaPicker];
+    [self dismissMediaPickerh];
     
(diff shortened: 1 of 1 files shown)


fridtjof:
`dismissMediaPickerh` looks like a typo :)

aaronraimist:
kind of, it was calling a different `dismissMediaPicker` so I just tacked on an extra letter so it would definitely call the same method. Didn't really look in to why that was happening but yeah we should use a better fix or rename that method so it makes more sense before it lands.

PR #559, opened 2019-05-29
MXRoomViewController: Refresh room bubbles cell data messages calculation on orientation change

SBiOSoftWhare:


MatrixKit/Models/Room/MXKRoomDataSource.m
```diff
@@ -1060,6 +1060,17 @@ - (CGFloat)cellHeightAtIndex:(NSInteger)index withMaximumWidth:(CGFloat)maxWidth
     return 0;
 }
 
+- (void)refreshBubblesCellData
```
giomfo: I don't think "refresh" is the right term here,
I would suggest "flushPrecomputedBubblesCellData"...

PR #561, opened 2019-06-06
MXKRoomDataSource: Add can react and can edit an event method

SBiOSoftWhare:


MatrixKit/Models/Room/MXKRoomDataSource.m
```diff
@@ -2132,6 +2132,16 @@ - (void)eventDidDecrypt:(NSNotification *)notif
     }
 }
 
+// Indicates whether an event has base requirements to allow actions (like reply, reactions, edit, etc.)
+- (BOOL)canPerformActionOnEvent:(MXEvent*)event
+{
+    BOOL isSent = event.sentState == MXEventSentStateSent;
+    BOOL isRoomMessage = event.sentState == MXEventSentStateSent;
```
giomfo: I think this boolean `isRoomMessage` is not correctly computed

PR #573, opened 2019-06-27
Reactions: Hide all reactions that aren't a single emoji

manuroe:
Fixes https://github.com/vector-im/riot-ios/issues/2508.

Note this PR is based on the `[MXKTools isSingleEmojiString:]` method that has a [bug](https://github.com/vector-im/riot-ios/issues/2479) with latest emojis but it works for the 8 emojis we use by default

### DIFF
--- a/MatrixKit/Categories/MXAggregatedReactions+MatrixKit.m
+++ b/MatrixKit/Categories/MXAggregatedReactions+MatrixKit.m
@@ -1,0 +1,44 @@
+/*
+ Copyright 2019 The Matrix.org Foundation C.I.C
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


MatrixKit/Categories/MXAggregatedReactions+MatrixKit.m
```diff
@@ -0,0 +1,44 @@
...
+
+#import "MXAggregatedReactions+MatrixKit.h"
+
+#import "MXKTools.h"
+
+@implementation MXAggregatedReactions (MatrixKit)
+
+- (nullable MXAggregatedReactions *)aggregatedReactionsWithSingleEmoji
+{
+    NSMutableArray *reactions = [NSMutableArray arrayWithCapacity:self.reactions.count];
```
SBiOSoftWhare: You can also use this syntax:
```
NSArray *singleEmojiReactions = [self.reactions filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(id object, NSDictionary *bindings) {
        MXReactionCount *reactionCount = (MXReactionCount*)object;
        return [MXKTools isSingleEmojiString:reactionCount.reaction]
    }]];
```

PR #592, opened 2019-08-21
Use MXIdentityService to perform identity server requests

SBiOSoftWhare:
vector-im/riot-ios/issues/2647

Requires matrix-org/matrix-ios-sdk/pull/706

### DIFF
--- a/MatrixKit/Models/MXK3PID.m
+++ b/MatrixKit/Models/MXK3PID.m
@@ -164,34 +164,51 @@
     // Sanity Check
     if (_validationState == MXK3PIDAuthStateTokenReceived)
     {
-        _validationState = MXK3PIDAuthStateTokenSubmitted;
-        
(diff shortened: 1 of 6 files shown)


MatrixKit/Models/MXK3PID.m
```diff
@@ -156,7 +166,7 @@ - (void)submitValidationToken:(NSString *)token
     {
         _validationState = MXK3PIDAuthStateTokenSubmitted;
         
-        currentRequest = [mxRestClient submit3PIDValidationToken:token medium:self.medium clientSecret:self.clientSecret sid:self.sid success:^{
+        currentRequest = [self.identityService submit3PIDValidationToken:token medium:self.medium clientSecret:self.clientSecret sid:self.sid success:^{
```
giomfo: call `failure` if `self.identityService` is null

PR #616, opened 2019-10-11
MXKContactManager: Limit the number of full lookups

manuroe:
Do it once per new matrix session.

The full lookup (done by updateMatrixIDsForAllLocalContacts) will be triggered by refreshMatrixContacts if needed.

Add an assert on updateMatrixIDsForAllLocalContacts to avoid to call it before data was loaded (by refreshMatrixContacts).


giomfo:
I would call `updateMatrixIDsForAllLocalContacts` when `discoverUsersBoundTo3PIDsBlock`is updated too. Wdyt?

manuroe:
> I would call updateMatrixIDsForAllLocalContacts when discoverUsersBoundTo3PIDsBlockis updated too. Wdyt?

IRL, we agreed that if you use `discoverUsersBoundTo3PIDsBlockis` you shortcut the use of IS. That means you do not care if the user changes their IS.

PR #621, opened 2019-10-21
Pasteboard: Fix a crash when passing a nil object

SBiOSoftWhare:
Passing nil to `-[UIPasteboard setString:]` causes a crash whereas `string` property is nullable (See Radar issue https://openradar.appspot.com/36063433).

### DIFF
--- a/MatrixKit/Controllers/MXKRoomViewController.m
+++ b/MatrixKit/Controllers/MXKRoomViewController.m
@@ -2830,8 +2830,16 @@
                                                                    // Cancel event highlighting
                                                                    [roomBubbleTableViewCell highlightTextMessageForEvent:nil];
                                                                    
-                                                                   [[UIPasteboard generalPasteboard] setString:selectedComponent.textMessage];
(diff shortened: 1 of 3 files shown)


MatrixKit/Controllers/MXKRoomViewController.m
```diff
@@ -2830,8 +2830,16 @@ - (void)dataSource:(MXKDataSource *)dataSource didRecognizeAction:(NSString *)ac
...
-                                                                   [[UIPasteboard generalPasteboard] setString:selectedComponent.textMessage];
+                                                                   NSString *textMessage = selectedComponent.textMessage;
                                                                    
+                                                                   if (textMessage)
+                                                                   {
+                                                                       [[UIPasteboard generalPasteboard] setString:textMessage];
+                                                                   }
+                                                                   else
+                                                                   {
+                                                                       NSLog(@"[MXKRoomViewController] Copy text failed. Text is nil.");
```
manuroe: Can we log the event id for further investigation?

PR #662, opened 2020-05-02
Add more flexibility to open links in other browsers, e.g. FF Mobile

### LINKED ISSUE #661: The http(s)LinkScheme setting is too simplistic
The MXKAppSetting properties httpLinkScheme and httpsLinkScheme allow for the use of a third-party browser to open links. This works great for Chrome but not for Firefox which expects a URL of the form firefox://open-url?url=_encodedUrl_ where encodedUrl is the target url encoded as query param (this is done by [addingPercentEncoding(withAllowedCharacters:](https://developer.apple.com/documentation/foundation/nsstring/1411946-addingpercentencoding) [urlQueryAllowed](https://developer.apple.com/documentation/foundation/nscharacterset/1416698-urlqueryallowed)).

This can become quite complex (as is at [Telegram](https://github.com/TelegramMessenger/Telegram-iOS/blob/master/submodules/OpenInExternalAppUI/Sources/OpenInOptions.swift)) because some browser expect a format like dolphin://http://whatever.com or yandexbrowser-open-url://escapedUrl where escapedUrl is url-host escaped.

While I think, the latter stuff is unneeded, the support for Firefox would be nice and is also [asked for at Riot](https://github.com/vector-im/riot-ios/issues/2949). I will work on this.

maxm123:
This PR adds two app settings - httpHttpsBaseURL and httpHttpsQueryParam. With their help, browsers like Firefox can be supported. If httpHttpsBaseURL is present and httpLinkScheme/httpsLinkScheme doesn't have its default value, the URL is constructed as follows: httpHttpsBaseURL?httpHttpsQueryParam=url.

An example would be FF with httpHttpsBaseURL=firefox://open-url and httpHttpsQueryParam=url. The escaping is handled by NSURLComponents as setQueryItems is used.

Shall I add some predefined settings for applications to use (such as FF, Chrome, etc.)?

Closes #661 
Signed-off-by: Maximilian Müller <maxm123@techie.com>


maxm123:
What is wrong with this PR? Should I do something so that it can be merged?

PR #674, opened 2020-06-02
Ensure Event Existence when Loading MXKRoomDataSource

ismailgulek:
Fix vector-im/riot-ios#3290

### DIFF
--- a/MatrixKit/Models/Room/MXKRoomDataSource.m
+++ b/MatrixKit/Models/Room/MXKRoomDataSource.m
@@ -228,15 +228,12 @@ NSString *const kMXKRoomDataSourceTimelineErrorErrorKey = @"kMXKRoomDataSourceTi
     }
     else
     {
-        //  wait for the specific event to be existent
+        //  give a chance for the specific event to be existent, for only one sync
         //  use kMXSessionDidSyncNotification here instead of MXSessionStateRunning, because session does not send this state update if it's already in this state
         id syncObserver = nil;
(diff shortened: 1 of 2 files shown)


MatrixKit/Models/Room/MXKRoomDataSource.m
```diff
@@ -167,41 +167,78 @@ @implementation MXKRoomDataSource
...
+    {
+        [self finalizeRoomDataSource:roomDataSource onComplete:onComplete];
+    }
+    else
+    {
+        //  wait for the specific event to be existent
+        //  use kMXSessionDidSyncNotification here instead of MXSessionStateRunning, because session does not send this state update if it's already in this state
+        id syncObserver = nil;
+        syncObserver = [[NSNotificationCenter defaultCenter] addObserverForName:kMXSessionDidSyncNotification object:mxSession queue:nil usingBlock:^(NSNotification * _Nonnull note) {
+            if ([mxSession.store eventExistsWithEventId:initialEventId inRoom:roomDataSource.roomId])
```
manuroe, line 235: The event may be never in the store.
If the event is old and the room very active, there will be a gap in the `/sync` that provides only the 10, 20, 30 last messages of a room.
We should give a chance during only one `/sync` and then leave.
If the event is not in the store, the code should be able to open the timeline around this event using `/context` around the event.

ismailgulek: Ok, updating.

PR #677, opened 2020-06-11
Riot 3337

ismailgulek:
Fix https://github.com/vector-im/riot-ios/issues/3337


ismailgulek:
Opened from wrong branch.

PR #680, opened 2020-06-29
Make sure PushKit Pusher removed before losing the Push Token

ismailgulek:
Part of vector-im/riot-ios#3369

### DIFF
--- a/MatrixKit/Models/Account/MXKAccountManager.m
+++ b/MatrixKit/Models/Account/MXKAccountManager.m
@@ -468,12 +468,13 @@ NSString *const kMXKAccountManagerDidSoftlogoutAccountNotification = @"kMXKAccou
             // turn off the Push flag for all accounts if any
             for (MXKAccount *account in mxAccounts)
             {
-                [account enablePushKitNotifications:NO success:nil failure:nil];
+                [account enablePushKitNotifications:NO success:^{
+                    //  make sure pusher really removed before losing token.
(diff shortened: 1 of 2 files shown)


MatrixKit/Models/Account/MXKAccountManager.m
```diff
@@ -468,12 +468,13 @@ - (void)setPushDeviceToken:(NSData *)pushDeviceToken withPushOptions:(NSDictiona
             // turn off the Push flag for all accounts if any
             for (MXKAccount *account in mxAccounts)
             {
-                [account enablePushKitNotifications:NO success:nil failure:nil];
+                [account enablePushKitNotifications:NO success:^{
+                    //  make sure pusher really removed before losing token.
+                    [[NSUserDefaults standardUserDefaults] removeObjectForKey:@"pushDeviceToken"];
+                    [[NSUserDefaults standardUserDefaults] removeObjectForKey:@"pushOptions"];
+                } failure:nil];
```
SBiOSoftWhare, line 475: We can add a log here for rageshakes investigation.

ismailgulek: It’s already logged at https://github.com/matrix-org/matrix-ios-kit/blob/1d11c01e97fdf89b5ed5ee457fce9a6552c3ef9a/MatrixKit/Models/Account/MXKAccount.m#L530

PR #710, opened 2020-08-28
Replace `cmark` with `Down`

ismailgulek:
Fixes vector-im/element-ios#3569


MatrixKit/Utils/EventFormatter/MarkdownParser.swift
```diff
@@ -0,0 +1,53 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import Foundation
+import Down
+import libcmark
+
+@objcMembers
+public class MarkdownParser: NSObject {
```
SBiOSoftWhare, line 24: In fact this is not a Markdown parser here but a Markdown to HTML renderer. So maybe a class name like `MarkdownToHTMLRenderer` is more appropriate.
To be able to have one renderer instance in `MXKEventFormatter` it can be better to remove the `public init(withMarkdown markdown: String)` method.

In that way we can create a `MarkdownToHTMLRendererProtocol` with this method: 
```
public func renderToHTML(markdown: String) -> String?
```
And use `@property (nonatomic, strong) id<MarkdownToHTMLRendererProtocol> markdownToHTMLRenderer` in `MXKEventFormatter`

MatrixKit/Utils/EventFormatter/MarkdownParser.swift
```diff
@@ -0,0 +1,53 @@
...
+        self.markdown = markdown
+        self.options = .default
+        super.init()
+    }
+    
+    public func toHTML() -> String? {
+        do {
+            return try Down(markdownString: markdown).toHTML(options)
+        } catch {
+            return nil
```
SBiOSoftWhare, line 39: It could be interesting to log error here but not leaking message, we can use "[MarkdownToHTMLRenderer] fail to render to HTML" for example. Otherwise prefer use `try?` in this case in order to return in one line: `return try? Down(markdownString: markdown).toHTML(options)`.

ismailgulek: Updated

MatrixKit/Utils/EventFormatter/MarkdownParser.swift
ismailgulek: Updated

MatrixKit/Utils/EventFormatter/MXKEventFormatter.m
```diff
@@ -1868,9 +1873,8 @@ - (UIFont*)fontForEvent:(MXEvent*)event
 
 - (NSString *)htmlStringFromMarkdownString:(NSString *)markdownString
 {
-    const char *cstr = [markdownString cStringUsingEncoding: NSUTF8StringEncoding];
-    const char *htmlCString = cmark_markdown_to_html(cstr, strlen(cstr), CMARK_OPT_HARDBREAKS);
-    NSString *htmlString = [[NSString alloc] initWithCString:htmlCString encoding:NSUTF8StringEncoding];
+    MarkdownParser *hardBreaksParser = [[MarkdownParserHardBreaks alloc] initWithMarkdown:markdownString];
```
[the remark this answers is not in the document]
ismailgulek: Updated

MatrixKit/Utils/EventFormatter/MXKEventFormatter.m
```diff
@@ -20,11 +20,16 @@
 
 @import MatrixSDK;
 @import DTCoreText;
-@import cmark;
 
 #import "MXEvent+MatrixKit.h"
 #import "NSBundle+MatrixKit.h"
 
+#if __has_include(<MatrixKit/MatrixKit-Swift.h>)
```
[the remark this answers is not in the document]
ismailgulek: I've updated the module name as `MatrixKit`.

MatrixKit/Utils/EventFormatter/MarkdownToHTMLRenderer.swift
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2015 OpenMarket Ltd
+Copyright 2017 Vector Creations Ltd
+Copyright 2018 New Vector Ltd
```
SBiOSoftWhare, line 4:
```suggestion
Copyright 2020 The Matrix.org Foundation C.I.C
```

SBiOSoftWhare (approved):

PR #736, opened 2020-10-13
Sync Response Handling and Background Sync Improvements

ismailgulek:
Part of vector-im/element-ios#3579


ismailgulek:
Not needed anymore.

PR #753, opened 2020-12-10
Contacts and Accounts encryption using keys provided by MXKeyProvider

gileluard:
fix https://github.com/vector-im/element-ios/issues/3866

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,7 @@ Changes to be released in next version
 =================================================
 
 ✨ Features
- * 
+ * Data encryption for MXKContactManager and MXKAccountManager using MXKeyProvider (#3866)
 
 🙌 Improvements
  * 
--- a/MatrixKit/Models/Account/MXKAccountManager.m
+++ b/MatrixKit/Models/Account/MXKAccountManager.m
@@ -28,6 +28,7 @@ static NSString *const kMXKAccountsKey = @"accountsV2";
 NSString *const kMXKAccountManagerDidAddAccountNotification = @"kMXKAccountManagerDidAddAccountNotification";
 NSString *const kMXKAccountManagerDidRemoveAccountNotification = @"kMXKAccountManagerDidRemoveAccountNotification";
 NSString *const kMXKAccountManagerDidSoftlogoutAccountNotification = @"kMXKAccountManagerDidSoftlogoutAccountNotification";
+NSString *const kMXKAccountManagerDataType = @"kMXKAccountManagerDataType";
 
 @interface MXKAccountManager()
 {
@@ -615,7 +616,8 @@ NSString *const kMXKAccountManagerDidSoftlogoutAccountNotification = @"kMXKAccou
         NSError *error = nil;
         NSData* filecontent = [NSData dataWithContentsOfFile:accountFile options:(NSDataReadingMappedAlways | NSDataReadingUncached) error:&error];
         
-        if (!error) {
+        if (!error)
+        {
             // Decrypt data if encryption method is provided
             filecontent = [self decryptData:filecontent];
             NSKeyedUnarchiver *decoder = [[NSKeyedUnarchiver alloc] initForReadingWithData:filecontent];
@@ -667,7 +669,7 @@ NSString *const kMXKAccountManagerDidSoftlogoutAccountNotification = @"kMXKAccou
 {
     // Exceptions are not cached as the key is always needed if the KeyProviderDelegate
     // is provided.
-    MXAesKeyData *keyData = (MXAesKeyData *) [[MXKeyProvider sharedInstance] requestKeyForDataOfType:kAccountType];
+    MXAesKeyData *keyData = (MXAesKeyData *) [[MXKeyProvider sharedInstance] requestKeyForDataOfType:kMXKAccountManagerDataType isMandatory:YES expectedKeyType:kAes];
     if (keyData)
     {
         NSData *cipher = [MXAes encrypt:data aesKey:keyData.key iv:keyData.iv error:nil];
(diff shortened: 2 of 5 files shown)


MatrixKit/Models/Account/MXKAccountManager.m
```diff
@@ -638,4 +665,51 @@ - (void)forceReloadAccounts
     [self loadAccounts];
 }
 
+- (NSData*)encryptData:(NSData*)data
+{
+    // Exceptions are not cached as the key is always needed if the KeyProviderDelegate
```
manuroe, line 670:
```suggestion
    // Exceptions are not caught as the key is always needed if the KeyProviderDelegate
```

CHANGES.rst
```diff
@@ -2,7 +2,7 @@ Changes to be released in next version
 =================================================
 
 ✨ Features
- * 
+ * Data encryption for MXKContactManager and MXKAccountManager using MXKeyProvider (#3866)
```
manuroe, line 5: We need to document in the API Changes section that the application must provide a `MXKeyProviderDelegate`.
Else, people will be surprised.

MatrixKit/Models/Account/MXKAccountManager.m
```diff
@@ -22,11 +22,13 @@
...
 #import "MXKTools.h"
 
-static NSString *const kMXKAccountsKey = @"accounts";
+static NSString *const kMXKAccountsKeyOld = @"accounts";
+static NSString *const kMXKAccountsKey = @"accountsV2";
 
 NSString *const kMXKAccountManagerDidAddAccountNotification = @"kMXKAccountManagerDidAddAccountNotification";
 NSString *const kMXKAccountManagerDidRemoveAccountNotification = @"kMXKAccountManagerDidRemoveAccountNotification";
 NSString *const kMXKAccountManagerDidSoftlogoutAccountNotification = @"kMXKAccountManagerDidSoftlogoutAccountNotification";
+NSString *const kMXKAccountManagerDataType = @"kMXKAccountManagerDataType";
```
manuroe, line 31:
```suggestion
NSString *const kMXKAccountManagerDataType = @"org.matrix.kit.MXKAccountManagerDataType";
```
in order to avoid collisions.

MatrixKit/Models/Account/MXKAccountManager.m
```diff
@@ -638,4 +665,51 @@ - (void)forceReloadAccounts
...
+- (void)migrateAccounts
+{
+    NSString *pathOld = [[MXKAppSettings cacheFolder] stringByAppendingPathComponent:kMXKAccountsKeyOld];
+    NSString *pathNew = [[MXKAppSettings cacheFolder] stringByAppendingPathComponent:kMXKAccountsKey];
+    NSFileManager *fileManager = [NSFileManager defaultManager];
+    if ([fileManager fileExistsAtPath:pathOld])
+    {
+        if (![fileManager fileExistsAtPath:pathNew])
+        {
+            mxAccounts = [NSKeyedUnarchiver unarchiveObjectWithFile:pathOld];
```
manuroe, line 707: Log this action please.

MatrixKit/Models/Account/MXKAccountManager.m
```diff
@@ -58,6 +60,9 @@ - (instancetype)init
     {
         _storeClass = [MXFileStore class];
         
+        // Migrate old account file to new format
+        [self migrateAccounts];
```
manuroe, line 64: In case of migration, the account will loaded twice but this is not an issue.

MatrixKit/Models/Contact/MXKContactManager.m
```diff
@@ -36,6 +37,8 @@
 
 NSString *const kMXKContactManagerDidInternationalizeNotification = @"kMXKContactManagerDidInternationalizeNotification";
 
+NSString *const kMXKContactManagerDataType = @"kMXKContactManagerDataType";
```
manuroe, line 40:
```suggestion
NSString *const kMXKContactManagerDataType = @"org.matrix.kit.MXKContactManagerDataType";
```

MatrixKit/Models/Contact/MXKContactManager.m
```diff
@@ -1833,17 +1836,19 @@ - (BOOL)encryptAndSaveData:(NSData*)data toFile:(NSString*)fileName
 
 - (NSData*)encryptData:(NSData*)data error:(NSError**)error fileName:(NSString*)fileName
 {
-    if (self.encryptionDelegate)
+    @try
     {
-        if (![self.encryptionDelegate isReady])
+        MXAesKeyData *keyData = (MXAesKeyData *) [[MXKeyProvider sharedInstance] requestKeyForDataOfType:kMXKContactManagerDataType isMandatory:NO expectedKeyType:kAes];
```
manuroe, line 1841: Can we use `isKindOfClass`?

gileluard: Actually because the expected key type is AES, the returned value will be necessary of type MXAesKeyData (otherwise an exception is raised).

I can add this check but it's unnecessary though

manuroe: I was thinking about compiler warning, like unsafe cast.
Perhaps the app build settings are less restrictive. So, I would prefer to fix it for the compiler, not the algo behind.

gileluard: Understood. Nevertheless, we have no warning as I explicitly cast the type of Key. Warnings are displayed only in case of implicit cast. Should I proceed ?

MatrixKit/Models/Account/MXKAccountManager.m
gileluard: I've tried to use `NSKeyedUnarchiver::decodeObject:` instead of `NSKeyedUnarchiver::decodeObjectForKey` so I could just read the data of the old file and then encrypt and save the result to the new file. `NSKeyedUnarchiver::decodeObject:` returns `nil` in this case unfortunately.

manuroe (approved):

PR #758, opened 2020-12-29
Fixed deep links to rooms from matrix.to no longer work

gileluard:
fix https://github.com/vector-im/element-ios/issues/3897

Actually it was still working but it was taking around 30s to navigate to the dedicated room. Links to same room was working and link to another room without event ID was working.

**Note:**
- Application doesn't navigate to the room if link is clicked from outside the application, deep link is not working (application is launch but no navigation)
- Not really sure why it's working if I set the eventID to nil but my assumption is that navigation to the dedicated room and finding the message with the right event ID are blocking each others.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Changes to be released in next version
  * 
 
 🐛 Bugfix
- * 
+ * MXKRoomDataSource: Fixed deep links to rooms from matrix.to no longer work
 
 ⚠️ API Changes
  * 
--- a/MatrixKit/Models/Room/MXKRoomDataSource.m
+++ b/MatrixKit/Models/Room/MXKRoomDataSource.m
@@ -200,7 +200,7 @@ NSString *const kMXKRoomDataSourceTimelineErrorErrorKey = @"kMXKRoomDataSourceTi
     //  In order to successfully fetch the room, we should wait for store to be ready.
     if (mxSession.state >= MXSessionStateStoreDataReady)
     {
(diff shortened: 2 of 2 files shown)


manuroe (commented):
Can you add an example of a public matrix.to link that does not work anymore to the issue?

It is weird to remove the event id here. I am afraid that the app will not be able to highlight the event in the the matrix.to link.

gileluard:
It's not about the link itself. As I wrote, the link will be opened after 30s if the application has to change room:

https://app.element.io/?pk_vid=9ff79ae5c9fa470216070270814d4bf8#/room/!mlObpGlNzvPSjyTRZX:matrix.org/$lrYMRcA1F-YVrVtlYM2mD_m-riZeBIp__aCkGYjyX2k?via=matrix.org&via=vector.modular.im&via=federator.dev
https://matrix.to/#/!mlObpGlNzvPSjyTRZX:matrix.org/$ZYx2Yyrtj7NGCeX11Uzbd2lB2rSB3utO5amFzthO43U?via=matrix.org&via=vector.modular.im&via=federator.dev

From a direct room for example is not behaving well.

gileluard:
Concerning the fix. It is weird to me as well but the navigation works well and I didn't notice any regression

manuroe:
The issue is probably around this but I would prefer us to clean code that it is no more useful.
We need to check the history behind.

Behavior has changed with recents changes on the codebase. It works better even if there is still sometimes this 30s bug.

PR #766, opened 2021-01-24
Don't query pasteboard contents until paste:

Adlai-Holler:
Currently `canPerformAction:` causes a read of the pasteboard, which is annoying on iOS 14. The system alerts the user "Pasted from [x]" every time the menu is queried.

Instead we can check the types without querying the contents.


Adlai-Holler:
@manuroe Updated with signoff & changelog.

PR #768, opened 2021-01-28
Remove iOS 7 auto layout code paths

Adlai-Holler:
Deployment target is iOS 9 now, so these can go.

Signed-off-by: Adlai Holler <adlai@icloud.com>


Adlai-Holler:
This one probably doesn't need a changelog entry, right?

PR #774, opened 2021-02-21
MXKRoomDataSource: Notify subclasses on room change

SBiOSoftWhare:


MatrixKit/Models/Room/MXKRoomDataSource.h
```diff
@@ -329,6 +329,11 @@ extern NSString *const kMXKRoomDataSourceTimelineErrorErrorKey;
  */
 - (void)reload;
 
+/**
+ Called when room property changed. Designed to be used by subclasses.
+ */
+- (void)roomDidChange;
```
manuroe, line 335: I am not confortable with `didChange` because it seems that it can be modified modified over time.
`didSet` may be a bit better.

manuroe (approved):

PR #789, opened 2021-03-23
Fix collapsing of separately processed events

Johennes:
As reported in vector-im/element-ios#4102, membership changes are sometimes wrongly collapsed. This happens when events of the same collapsible series are processed in different batches. For forward directed events, only the start of a collapsible series is tracked in `collapsingCellDataSeriess`. As a result, when the event that corresponds to the beginning of the series was processed in a preceding event batch, the series will be missing in `collapsingCellDataSeriess` and its summary string won't get updated.

This commit fixes the situation by adding the start of the series to `collapsingCellDataSeriess` when it is missing and the series is appended to.

Signed-off-by: Johannes Marbach <n0-0ne+github@mailbox.org>

### Pull Request Checklist

* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Changes to be released in next version
  * Pods: Update JSQMessagesViewController, DTCoreText, Down (vector-im/element-ios/issues/4120).
 
 🐛 Bugfix
- * 
+ * Fix collapsing of separately processed events
 
 ⚠️ API Changes
  * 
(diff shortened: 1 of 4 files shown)


manuroe (approved):
Thanks a lot @Johennes for the fix and unit tests you added for it.

PR #793, opened 2021-03-25
MXKRoomBubbleCellDataStoring: Introduce target user ID, display name and avatar URL for room membership events

Johennes:
As reported in vector-im/element-ios#4102, collapsed membership changes in the Element iOS app currently _always_ display the sender avatar. In contrast, the Element web app shows the avatar of the user whose membership has changed in this case.

This commit introduces new properties `targetId`, `targetDisplayName`, `targetAvatarUrl` and `targetAvatarPlaceholder` on `MXKRoomBubbleCellDataStoring`. The values are populated from the event (in case it is an `m.room.member` event) in the initializer of `MXKRoomBubbleCellData` (where the corresponding sender properties are already being assigned). This way, clients can easily choose to present the target avatar instead of the sender avatar when processing the bubble data.

Note that the "target" terminology was taken over from matrix-js-sdk which populates the `target` property on `MatrixEvent` objects when they are added to the timeline.

Signed-off-by: Johannes Marbach <n0-0ne+github@mailbox.org>

### Pull Request Checklist

* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,7 @@ Changes to be released in next version
  * 
 
 ⚠️ API Changes
- * 
+ * MXKRoomBubbleCellDataStoring: Introduce target user ID, display name and avatar URL for room membership events (vector-im/element-ios/issues/4102).
 
 🗣 Translations
  * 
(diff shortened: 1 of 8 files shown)


manuroe:
This is awesome. Thanks!

PR #840, opened 2021-06-25
CallVC: Audio Routing Menu

ismailgulek:
Requires https://github.com/matrix-org/matrix-ios-sdk/pull/1138

### DIFF
--- a/MatrixKit/Controllers/MXKCallViewController.m
+++ b/MatrixKit/Controllers/MXKCallViewController.m
@@ -822,67 +822,58 @@ static const CGFloat kLocalPreviewMargin = 20;
 
 - (void)showAudioDeviceOptions
 {
-    //  create the alert
-    currentAlert = [UIAlertController alertControllerWithTitle:nil
-                                                       message:nil
-                                                preferredStyle:UIAlertControllerStyleActionSheet];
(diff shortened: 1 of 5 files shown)


MatrixKit/Controllers/MXKCallViewController.m
```diff
@@ -808,51 +822,58 @@ - (IBAction)onButtonPressed:(id)sender
...
-        MXStrongifyAndReturnIfNil(self);
-        self->currentAlert = nil;
-        self->mxCall.audioToSpeaker = NO;
+    for (MXAudioOutputRoute *route in availableRoutes)
+    {
+        //  route action
+        NSString *name = route.name;
+        if (route.routeType == MXAudioOutputRouteTypeLoudSpeakers)
+        {
+            name = [NSBundle mxk_localizedStringForKey:@"call_more_actions_audio_use_device"];
```
manuroe, line 834: I wonder if we should have a localisation string for every route type so that the UI will not depend on the system and it will be more robust and predictive.

manuroe: answer: no. `route.name` can be the real device name. This is useless to try to localise it...

manuroe (approved):

PR #846, opened 2021-07-05
Tidy up following on from element-ios#4530

pixlwave:
Following on from https://github.com/vector-im/element-ios/pull/4530 this removes `showDecryptedContentInNotifications` from `MXKAccount` as it is no longer used.


langleyd (approved):
LGTM pending design/product sign off on the changed behaviour.

PR #850, opened 2021-07-13
vector-im/element-ios/issues/4090 - Add custom reply text to voice me…

stefanceriu:
…ssages, add duration and samples to sendVoiceMessage.


pixlwave (approved):
LGTM.

PR #852, opened 2021-07-14
Fix timeline scroll position when viewing a room for the second time.

pixlwave:
Fixes https://github.com/vector-im/element-ios/issues/4524.

This was a regression made [here](https://github.com/matrix-org/matrix-ios-kit/pull/831/commits/f24c00cd113154a3d19a4cc3bddfb67e3f4d5729) in v0.15.1. Needs testing against a virtual timeline to confirm it hasn't undone what it fixed in that situation.


manuroe:
The change looks good to me. I will let @ismailgulek check it in a virtual timeline context.

ismailgulek (approved):

PR #854, opened 2021-07-15
Failed voice and audio message sending retries

stefanceriu:
Implemented retrying for failed audio file and voice messages always calling completion blocks.


pixlwave (approved):
Looks good to me. Just needs a CHANGES.rst?

PR #857, opened 2021-07-15
Standardise casing of identity server

psrpinto:
This PR standardises the spelling and casing of the following terms, across the codebase, as per vector-im/element-ios#2655:

- homeserver
- identity server
- integration manager

While the issue mentions only user-visible text, this PR changes all instances of the terms, regardless of whether they're user-visible or not. I think there's value in standardising these terms even in non-user-visible text, so I went ahead and applied the changes to all text.

For reference, here's the script I used to find candidates for replacement: https://gist.github.com/psrpinto/b3787bae212d5d99649b517e2efd4dce

Related to vector-im/element-ios#4559


psrpinto:
@manuroe Any chance we could get this one merged? Should be riskless.

manuroe:
Thanks @psrpinto for cleaning up this and for sharing your useful script!

PR #878, opened 2021-08-12
Lazy load room messages

ismailgulek:
Part of vector-im/element-ios#4382


ismailgulek:
Replaced by https://github.com/matrix-org/matrix-ios-kit/pull/883

PR #889, opened 2021-09-01
vector-im/element-ios/issues/4693 - Drop iOS 11 and lower support.

stefanceriu:


langleyd (approved):
This LGTM. Looks like a fun PR, just deleting code. 😂

PR #893, opened 2021-09-07
Add link property to MXKRoomBubbleComponent and parse events for links.

pixlwave:
As part of https://github.com/vector-im/element-ios/pull/4790, the change to MatrixKit for URL previews include the following:
- A `link` property on `MXKRoomBubbleComponent`.
- Link detection on creation and update of `MXKRoomBubbleComponent`.
- Links are only detected for unencrypted message events.
- Link detection is controlled by an `MXKAppSettings` which is disabled by default (but enabled in the Element PR).
- Fixes a crash that could occur reloading the timeline data whilst also rotating the device.


MatrixKit/Models/Room/MXKRoomBubbleCellData.m
```diff
@@ -587,6 +587,21 @@ - (BOOL)hasAttributedTextMessage
     return hasAttributedTextMessage;
 }
 
+- (BOOL)hasLink
+{
+    @synchronized (bubbleComponents) {
+        for (MXKRoomBubbleComponent *component in bubbleComponents)
```
pixlwave, line 593: In Element we break up the bubbleComponents so each cell data object has only 1 link and it is always from the last component in the data: 
An optimisation here could be to iterate through the components backwards or possibly to look directly at the last component (although this behaviour would be poor for other clients using the framework).

MatrixKit/Categories/NSString+MatrixKit.swift
```diff
@@ -0,0 +1,58 @@
...
+            guard let url = NSURL(string: urlString) else { return }
+            
+            // ensure the match is a web link
+            guard let scheme = url.scheme?.lowercased(),
+                  scheme == "https" || scheme == "http"
+            else { return }
+            
+            // discard any matrix.to links
+            guard let host = url.host?.lowercased(),
+                  host != "matrix.to"
```
SBiOSoftWhare, line 49: We should had or use constants like `kMXMatrixDotToUrl`. But probably better to add a parameters like `hostsToIgnore: [String]`. From the app you can decide which URLs to ignore. A default implementation can use something like: `hostsToIgnore: [URL(string: kMXMatrixDotToUrl.host)]` (make null safety checks before).

pixlwave: Updated, which makes the function more generally usable, although I just realised doesn't allow customisation in `MXKRoomBubbleComponent`'s [detection](https://github.com/matrix-org/matrix-ios-kit/pull/893/files#diff-b320829d3dedae327a784fc36287f2698f84ed60639b987ce124aea30f26c7e7R128).

Wondering if it would make more sense to add a `firstURLDetectionIgnoredHosts` setting to `MXKAppSetting` and use that when no arguments are passed to `mxk_firstURLDetected()`.

SBiOSoftWhare: It worth to keep the method with open parameters for test and reuse. We can keep `firstURLDetectionIgnoredHosts` as is and `mxk_firstURLDetected()` can use `MXKAppSetting.urlPreviewIgnoredHosts` for example

pixlwave: Updated, I think it should make sense.

MatrixKit/Models/MXKAppSettings.m
```diff
@@ -491,6 +493,30 @@ - (void)setEnableBubbleComponentLinkDetection:(BOOL)storeLinksInBubbleComponents
     }
 }
 
+- (NSArray<NSString *> *)firstURLDetectionIgnoredHosts
+{
+    if (self == [MXKAppSettings standardAppSettings])
+    {
+        return [NSUserDefaults.standardUserDefaults objectForKey:@"firstURLDetectionIgnoredHosts"] ?: @[[NSURL URLWithString:kMXMatrixDotToUrl].host];
```
SBiOSoftWhare, line 500: We can remove `?: @[[NSURL URLWithString:kMXMatrixDotToUrl].host];` as if someone wants to force all linked to be detected he can. `firstURLDetectionIgnoredHosts = nil` or `= @[]` can be accepted.

SBiOSoftWhare (approved):

PR #901, opened 2021-09-15
Exposed more room joining result types in the completion handler.

stefanceriu:
* exposed various room joining results
* converged room joining failure handlers
* bailing our early when possible


langleyd (approved):
LGTM

PR #904, opened 2021-09-15
Revert removal of default ignored hosts value.

pixlwave:
Fixes https://github.com/vector-im/element-ios/issues/4826. The matrix.to host was only stored in the synthesized variable and not the user defaults.

This returns matrix.to when the user defaults return `nil` and always sets `@[]` instead of `nil`, to match the desirable behaviour discussed [here](https://github.com/matrix-org/matrix-ios-kit/pull/893#discussion_r704513073).

### DIFF
--- a/MatrixKit/Models/MXKAppSettings.m
+++ b/MatrixKit/Models/MXKAppSettings.m
@@ -511,7 +511,7 @@ static NSString *const kMXAppGroupID = @"group.org.matrix";
     {
         if (ignoredHosts == nil)
         {
-            ignoredHosts = @[]
+            ignoredHosts = @[];
         }
         
         [NSUserDefaults.standardUserDefaults setObject:ignoredHosts forKey:@"firstURLDetectionIgnoredHosts"];
--- a/MatrixKit/Models/MXKAppSettings.m
+++ b/MatrixKit/Models/MXKAppSettings.m
@@ -497,7 +497,7 @@ static NSString *const kMXAppGroupID = @"group.org.matrix";
 {
     if (self == [MXKAppSettings standardAppSettings])
     {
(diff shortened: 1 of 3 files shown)


MatrixKit/Models/MXKAppSettings.m
```diff
@@ -497,7 +497,7 @@ - (void)setEnableBubbleComponentLinkDetection:(BOOL)storeLinksInBubbleComponents
 {
     if (self == [MXKAppSettings standardAppSettings])
     {
-        return [NSUserDefaults.standardUserDefaults objectForKey:@"firstURLDetectionIgnoredHosts"];
+        return [NSUserDefaults.standardUserDefaults objectForKey:@"firstURLDetectionIgnoredHosts"] ?: @[[NSURL URLWithString:kMXMatrixDotToUrl].host];
```
SBiOSoftWhare, line 500: If we are registering explicitly an empty array or nil via `setFirstURLDetectionIgnoredHosts` we will always have `@[[NSURL URLWithString:kMXMatrixDotToUrl].host` as result. The question is why we need to do this, do we have an issue with user defaults reset?

MatrixKit/Models/MXKAppSettings.m
```diff
@@ -509,6 +509,11 @@ - (void)setFirstURLDetectionIgnoredHosts:(NSArray<NSString *> *)ignoredHosts
 {
     if (self == [MXKAppSettings standardAppSettings])
     {
+        if (ignoredHosts == nil)
+        {
+            ignoredHosts = @[]
```
SBiOSoftWhare, line 514:
```suggestion
            ignoredHosts = @[];
```

SBiOSoftWhare (approved):

PR #915, opened 2021-09-27
Simplify last message types

pixlwave:
Fixes https://github.com/vector-im/element-ios/issues/4451. Depends on https://github.com/matrix-org/matrix-ios-sdk/pull/1240.

Adds an `eventsFilterForLastMessage` app setting which defaults to only show Messages, Call Invite/Answer/Hangup and Stickers which matches the behaviour on Web. (Web also has reactions in its preview type list, but that is disabled via a setting elsewhere).


stefanceriu (approved):
Had a separate conversation about this, otherwise lgtm 👍

stefanceriu (approved):

PR #925, opened 2021-10-05
Add an analytics value for accept/decline of contacts access

pixlwave:
Part of vector-im/element-ios#4955, this PR adds an analytics value of accept/decline of contacts access.

### DIFF
--- a/MatrixKit/Utils/MXKConstants.h
+++ b/MatrixKit/Utils/MXKConstants.h
@@ -37,3 +37,15 @@ FOUNDATION_EXPORT NSString *const kMXKErrorNotification;
  */
 FOUNDATION_EXPORT NSString *const kMXKErrorUserIdKey;
 
+
+#pragma mark - Analytics
+
+/**
+ The analytics category for local contacts.
+ */
+FOUNDATION_EXPORT NSString *const kMXKAnalyticsContactsCategory;
+
+/**
+ The analytics value for accept/decline of local contacts access.
+ */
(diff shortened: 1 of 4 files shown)


MatrixKit/Utils/MXKConstants.h
```diff
@@ -37,3 +37,15 @@ FOUNDATION_EXPORT NSString *const kMXKErrorNotification;
  */
 FOUNDATION_EXPORT NSString *const kMXKErrorUserIdKey;
 
+
+#pragma mark - Analytics
+
+/**
+ The analytics category for local contacts.
+ */
+FOUNDATION_EXPORT NSString *const kMXKAnalyticsContactsCategory;
```
SBiOSoftWhare, line 46: Prefix `k` convention is normally used for internal constants. As this is a new constant we can remove it.
We can also use a dedicated `MXKAnalyticsCategories.h` or `MXKAnalyticsConstants.h`.

MatrixKit/Utils/MXKConstants.h
```diff
@@ -37,3 +37,15 @@ FOUNDATION_EXPORT NSString *const kMXKErrorNotification;
...
+
+/**
+ The analytics value for accept/decline of local contacts access.
+ */
+FOUNDATION_EXPORT NSString *const kMXKAnalyticsContactsAccessGranted;
```
SBiOSoftWhare, line 51: Same remark about `k` prefix.

PR #929, opened 2021-10-12
vector-im/element-ios/issues/4899 - MXKEventFormatter: Fixed invalid …

stefanceriu:
…parameters passed to generated localization functions.

### DIFF
--- a/MatrixKit/Utils/EventFormatter/MXKEventFormatter.m
+++ b/MatrixKit/Utils/EventFormatter/MXKEventFormatter.m
@@ -997,11 +997,11 @@ static NSString *const kHTMLATagRegexPattern = @"<a href=\"(.*?)\">([^<]*)</a>";
             {
                 if (isRoomDirect)
                 {
-                    displayText = [MatrixKitL10n noticeRoomAliasesForDm:aliases];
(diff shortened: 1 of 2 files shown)


MatrixKit/Utils/EventFormatter/MXKEventFormatter.m
```diff
@@ -997,11 +997,11 @@ - (NSAttributedString *)attributedStringFromEvent:(MXEvent *)event withRoomState
             {
                 if (isRoomDirect)
                 {
-                    displayText = [MatrixKitL10n noticeRoomAliasesForDm:aliases];
+                    displayText = [MatrixKitL10n noticeRoomAliasesForDm:aliases.description];
```
manuroe, line 1000: I wonder if it would be better to use `[aliases componentsJoinedByString:@", "] to have a more predictable behavior?
Do you have an example of the rendering?

pixlwave: FWIW, there's a new `NSListFormatter` for iOS 13+ that provides locale aware lists that might be worth looking at too.

manuroe (approved):

PR #934, opened 2021-10-18
Fix crash when pasting images backed by NSData.

pixlwave:
Kit part of https://github.com/vector-im/element-ios/issues/2076. When pasting WebP images from Safari, they're backed back `NSData` instead of `UIImage` and this wasn't handled.


stefanceriu (approved):
👍

PR #940, opened 2021-10-22
Threads P0

ismailgulek:
This PR makes threaded events look like regular replies.

`<blockquote>` css does not handle background color for these _replies_, i'll address this issue later.

Fixes vector-im/element-ios#5007

Requires https://github.com/matrix-org/matrix-ios-sdk/pull/1272


stefanceriu (approved):
👍

PR #944, opened 2021-11-01
Keep strong refs to views before display.

pixlwave:
Part of the fix for https://github.com/vector-im/element-ios/issues/5071.


stefanceriu (approved):
👍

PR #950, opened 2021-11-11
Fix unresponsive "Jump to unread"

langleyd:
Fixes https://github.com/vector-im/element-ios/issues/4071

I think we can remove this check altogether @ismailgulek. 

It was introduced in an [attempt to fix an issue](https://github.com/matrix-org/matrix-ios-kit/pull/674) with notifications not going to the message. But we no longer [highlight messages from notifications](https://github.com/vector-im/element-ios/pull/3341).

In the codes current state, I don't think this check is having any effect(maybe it did in the past) other than causing a delay in the join.

At the moment it's just called for the "jump to unread" button and when joining a room with an initialEventId, in both cases we are starting the timeline at a point in the past.

When we start the timeline at an event in the past [we load an in memory event store](https://github.com/matrix-org/matrix-ios-sdk/blob/develop/MatrixSDK/Data/MXEventTimeline.m#L89) rather than the persisted session store.

So from my testing the [initial check](https://github.com/matrix-org/matrix-ios-kit/blob/develop/MatrixKit/Models/Room/MXKRoomDataSource.m#L269) for the event in this in memory store always fails. We then wait for the next sync, but don't make any requests that would trigger the server to return. This means it hangs for an arbitrary amount of time(0-15 seconds for me), until the server pushes something on sync. During this time the join/jump to unread appear unresponsive.

Then we finalise the room and load the event using `/context`.

### DIFF
--- a/MatrixKit/Models/Room/MXKRoomDataSource.m
+++ b/MatrixKit/Models/Room/MXKRoomDataSource.m
@@ -234,7 +234,7 @@ typedef NS_ENUM (NSUInteger, MXKRoomDataSourceError) {
     //  In order to successfully fetch the room, we should wait for store to be ready.
     if (mxSession.state >= MXSessionStateStoreDataReady)
     {
-        [self ensureInitialEventExistenceForDataSource:roomDataSource initialEventId:initialEventId andMatrixSession:mxSession onComplete:onComplete];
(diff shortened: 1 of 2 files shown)


ismailgulek (approved):

langleyd:
Just to clarify one point, the check failed as past events are stored in the in memory store and while new events go to the session store.

PR #957, opened 2021-11-26
Various poll tweaks

stefanceriu:
* added poll start event type
* do not merge poll start events and allow actions on them.
* added unstable poll prefixes to the allow lists.


stefanceriu:
No longer needed here after migrating to Element.

